PR #4, opened 2018-12-20
Adds a note in case someone is unfamiliar with cocoapods.

drob:


GrandPoohBear:
Nevermind - I'm a dork.  Not as familiar with the fork merging stuff...

PR #9, opened 2018-12-21
Add typings for Typescript

pachuka:
Based this off of the documentation at: https://docs.heapanalytics.com/reference

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,114 @@
+/**
+ * Sets the app ID where you'll be sending data. It can be used to switch between projects or
+ * between your production and development environments.
+ *
+ * `setAppId` initializes Heap tracking on iOS. It should be called only once per app open.
+ * If it is called multiple times, it will initiate a new session for the user.
+ *
+ * @param appId the appId corresponding to one of your projects
+ */
+export function setAppId(appId: string): void;
+
+/**
+ * Send additional events to Heap.
+ *
+ * For Android, the properties map may be null to track an event with no properties.
+ * The property map keys and values must be converted to strings, but they can be queried as numbers in Heap.
+ *
+ * Custom event properties can be queried in the same fashion as any other event property.
+ *
+ * @param event name of the custom interaction. Limited to 255 characters.
+ * @param withProperties a JSON object containing key-value paiurs to be associated with an event. Keys and values
+ * must be a number or string.
+ */
+export function track(event: string, withProperties: object): void;
+
+/**
+ * Attach a unique identity to a user.
+ *
+ * If you assign the same identity to a user on a separate device, their past sessions and event activity will be
+ * merged into the existing Heap user with that identity.
+ *
+ * To better understand the client-side `identify` and `addUserProperties` APIs,
+ * [take a look at our comprehensive guide.](https://docs.heapanalytics.com/docs/using-identify)
+ *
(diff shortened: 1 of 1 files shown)


drob:
This is awesome, thank you for the PR! 🎉🎉🎉

We'll take a look promptly.

index.d.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * Sets the app ID where you'll be sending data. It can be used to switch between projects or
+ * between your production and development environments.
+ *
+ * `setAppId` initializes Heap tracking on iOS. It should be called only once per app open.
+ * If it is called multiple times, it will initiate a new session for the user.
```
drob: If it's called more than once in a react native iOS app, subsequent calls will be ignored. @GrandPoohBear is that the behavior on android or will it create a new session?

index.d.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @param appId the appId corresponding to one of your projects
+ */
+export function setAppId(appId: string): void;
+
+/**
+ * Send additional events to Heap.
+ *
+ * For Android, the properties map may be null to track an event with no properties.
```
drob: Is this specific to android?

index.d.ts
```diff
@@ -0,0 +1,114 @@
...
+export function addUserProperties(properties: object): void;
+
+/**
+ * Specify a set of global key-value pairs to get attached to all of a user's subsequent events.
+ *
+ * These event properties will persist across multiple sessions on the same device and get applied to both
+ * auto-captured and custom events.
+ *
+ * This is useful if you have some persistent state, but you don't want to apply it across all of a user's
+ * events with `identify`. A good example is "Logged In", which changes over the user's lifecycle.
```
drob: Should be `addUserProperties` here, not `identify`.

index.d.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+export function removeEventProperty(property: string): void;
+
+/**
+ * Removes all stored event properties.
+ */
+export function clearEventProperties(): void;
+
+/**
+ * Changes the frequency with which events are sent back to the Heap server from an iOS mobile device.
```
drob: I think this is a copy-paste mistake. I don't think we expose `changeInterval` yet.

index.d.ts
GrandPoohBear: Yeah, unfortunately this behavior is different across platforms right now, but it's something we will have to address from the native lib side.  Current behavior as of 0.1.0:
iOS - Subsequent calls will be ignored, so only the first appId will ever be used (by default, the native lib stops sending events if this gets called twice, so the bridge prevents it.)
Android - This initializes an entirely new tracker object, so the session will be created from scratch.

index.d.ts
GrandPoohBear: The old React Native library had exposed it on iOS, but Android only had it as a private function (IIRC).  To keep the same interface, we dropped that from the initial version.

pachuka: My mistake, just removed it.

index.d.ts
pachuka: I updated it with a note about the differences between behavior on ios vs android.

index.d.ts
pachuka: Removed the 'For Android, '

PR #13, opened 2019-01-11
Update README to include details about updating Podfile for React dependencies

jnatkins:
Our default podspec pulls in React 0.11, which can cause build failures if React isn't already overridden in the target project's Podfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,24 @@ Add the following to your Podfile:
 pod "react-native-heap", path: "../node_modules/@heap/react-native-heap"
 ```
 
+If your Podfile does not specify the path to your local React pod, you should add that (and any other necessary subspecs) to your Podfile. For example:
+
+```ruby
+pod 'React', :path => '../node_modules/react-native', :subspecs => [
+  'Core',
+  'CxxBridge', # Include this for RN >= 0.47
+  'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,24 @@ Add the following to your Podfile:
 pod "react-native-heap", path: "../node_modules/@heap/react-native-heap"
 ```
 
+If your Podfile does not specify the path to your local React pod, you should add that (and any other necessary subspecs) to your Podfile. For example:
```
GrandPoohBear: Was this drawn from the docs page at the link on line 19?  If so, we should probably put them closer together, or possibly just direct the user to that page more strongly.

One other idea - maybe we could have a `Build Troubleshooting` section of the docs where we include the error that people saw and how that might fix it.  Thoughts?

jnatkins: I pulled it from a Podfile I was working with, so not sure what docs page you're referring to, but would be happy to link to one. Can you share the link you're suggesting?

But yes, a `Build Troubleshooting` section sounds like a better idea than putting this directly in there. Share the link, and I'll handle both updates.

GrandPoohBear: This is the link from the RN docs: https://facebook.github.io/react-native/docs/integration-with-existing-apps#configuring-cocoapods-dependencies

jnatkins:
Updated based on feedback from @GrandPoohBear

PR #16, opened 2019-01-18
Autotrack touchable press+longPress with babel

jmtaber129:
This PR was originally just a POC for autocapture, but ended up turning into the feature branch for the initial `Touchable` autotrack functionality w/ prop capture to be released.

**Original description:**
This PR includes two parts:
* the instrumentation babel plugin (with tests!)
* the autotrack handler

The babel plugin instruments the `touchableHandlePress` and `touchableHandleLongPress` handlers on components that are both created with `createReactClass` and include `Touchable.Mixin` in its mixins.

The autotrack handler takes the event type (e.g. `touchableHandlePress`), the `Touchable` component's `this` reference, and the event passed to the press handler.  The handler extracts the component hierarchy + other event props, and sends a manual track with the resultant event.


js/Heap.js
```diff
@@ -22,28 +22,75 @@ export default Package.create({
...
+    return '';
+  }
+
+  // Skip components we don't care about.
+  // :TODO: (jmtaber129): Skip components with names/display names like 'View' and '_class'.
+  if (!(currNode.elementType !== 'RCTView' && currNode.elementType !== null && (currNode.elementType.displayName || currNode.elementType.name))) {
+    return `${getComponentHierarchy(currNode.return)}`;
+  }
+
+  const elementName = currNode.elementType.displayName || currNode.elementType.name || null;
```
GrandPoohBear: Can probably lose the null - we've already checked above that one of these is populated.

js/Heap.js
```diff
@@ -22,28 +22,75 @@ export default Package.create({
...
+
+  // If the element is a button, capture its props.
+  // :TODO: (jmtaber129): Change this once we allow configurably captured props.
+  let propsString = '';
+  if (elementName === 'Button') {
+    const props = currNode.stateNode.props;
+    const keys = Object.keys(props);
+    keys.forEach((key) => {
+      if (Object(props[key]) !== props[key]) {
```
GrandPoohBear: `if (props[key] !== null && props[key] !== undefined && typeof props[key] !== 'function' && typeof props[key] !== 'object')`

package.json
```diff
@@ -18,19 +18,27 @@
     "android",
     "ios",
     "js",
+    "babel_plugin",
```
GrandPoohBear: should change here too

package.json
```diff
@@ -18,19 +18,27 @@
...
   "publishConfig": {
     "access": "public"
   },
   "dependencies": {
+    "@babel/core": "^7.2.2",
+    "babel-template": "^6.26.0",
+    "babel-types": "^6.26.0",
     "flat": "^4.1.0",
-    "react-native-package": "^2.5.0"
+    "mocha": "^5.2.0",
```
GrandPoohBear: mocha and should should go into devDependencies

instrumentor/src/index.js
```diff
@@ -0,0 +1,101 @@
+// :TODO: (jmtaber129): Add file-level comment explaining what this plugin does and how it works.
```
GrandPoohBear: Yes - this is worth revisiting soon before the details fade away!

js/Heap.js
```diff
@@ -1,49 +1,128 @@
...
+  // Events
+  track: track,
+
+  // Redux middleware
+  reduxMiddleware: store => next => action => {
+    RNHeap.track('Redux Action', flatten(action));
+    next(action);
+  },
+
+  autotrackPress: (eventType, componentThis, event) => {
```
GrandPoohBear: I wonder if we ought to move this whole function to its own file, just to keep this one nice and neat.

jmtaber129:
:ship:ing it

PR #19, opened 2019-01-31
 Adding a prop extractor configurable with a built-in config

GrandPoohBear:


js/Heap.js
```diff
@@ -68,47 +70,41 @@ export default Package.create({
   },
 });
 
-const getComponentHierarchy = (componentThis) => {
```
GrandPoohBear: This is all just `prettier` formatting until we get down below...

js/Heap.js
```diff
@@ -68,47 +70,41 @@ export default Package.create({
...
-    || !(currNode.elementType.displayName || currNode.elementType.name)) {
+  if (
+    currNode.elementType === 'RCTView' ||
+    currNode.elementType === null ||
+    !(currNode.elementType.displayName || currNode.elementType.name)
+  ) {
     return getFiberNodeComponentHierarchy(currNode.return);
   }
 
-  const elementName = currNode.elementType.displayName || currNode.elementType.name;
```
GrandPoohBear: This is the actual new stuff in this file.

js/util/__tests__/extractProps.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    const obj2 = _.merge({}, obj1, {
+      props: {
+        a: 'bracket][bracket][bracket]',
+      },
+    });
+
+    expect(extractProps('Element', obj2, config)).toEqual(
+      '[a=bracketbracketbracket];[c=true];'
+    );
+  });
```
jmtaber129: Should add a test when a prop is an array.

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
+const pick = require('lodash.pick');
+const flatten = require('flat');
+
+export interface Component {
+  //TODO: use the real RN ones
```
jmtaber129: TODOs should be of the form `// :TODO: (<username of person with the most context (not necessarily author)>): <todo comment>`

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
+const pick = require('lodash.pick');
+const flatten = require('flat');
+
+export interface Component {
```
jmtaber129: Is there a `Component` type we can import directly from React?

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface PropExtractorConfig {
+  [componentName: string]: PropExtractorCriteria;
+}
+
+export const extractProps = (
+  elementName: string,
+  component: Component,
+  config: PropExtractorConfig
+): string => {
+  if (!component || !config[elementName]) {
```
jmtaber129: Do we need the `!component`? I'd think that wouldn't be necessary unless `component` was of type `Component | null` or `Component?`.

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
...
+  const filteredProps = pick(component.props, config[elementName].include);
+  const flattenedProps = flatten(filteredProps);
+  let propsString = '';
+
+  // Only include props that are primitives.
+  Object.keys(flattenedProps).forEach(key => {
+    if (
+      flattenedProps[key] !== null &&
+      flattenedProps[key] !== undefined &&
+      typeof flattenedProps[key] !== 'function'
```
jmtaber129: Should exclude arrays, too

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
...
+    ) {
+      // Remove all brackets from string
+      let prop = flattenedProps[key].toString().replace(/[\[\]]/g, '');
```
jmtaber129: const

js/Heap.js
```diff
@@ -1,6 +1,8 @@
 // Libraries
 import { NativeModules, Platform } from 'react-native';
 import Package from 'react-native-package';
+import { extractProps } from './util/extractProps';
+import { builtinPropExtractorConfig } from './propExtractorConfig';
 
 let flatten = require('flat');
```
jmtaber129: Oof, this should be const

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
...
+  let propsString = '';
+
+  // Only include props that are primitives.
+  Object.keys(flattenedProps).forEach(key => {
+    if (
+      flattenedProps[key] !== null &&
+      flattenedProps[key] !== undefined &&
+      typeof flattenedProps[key] !== 'function'
+    ) {
+      // Remove all brackets from string
```
jmtaber129: Nit: period at end of comment

js/util/extractProps.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const filteredProps = pick(component.props, config[elementName].include);
+  const flattenedProps = flatten(filteredProps);
+  let propsString = '';
+
+  // Only include props that are primitives.
+  Object.keys(flattenedProps).forEach(key => {
+    if (
+      flattenedProps[key] !== null &&
```
jmtaber129: I guess this is because of the prettier config, but in the case of binary operations split over multiple lines, I generally prefer the operator to be on the new line, e.g.
```
if (flattendProps[key] !== null
    && flattendProps[key] !== undefined)
```

I'll try to see if we have any guidance on this sort of thing in any internal style guides, but I don't think we do :/

js/Heap.js
```diff
@@ -1,6 +1,8 @@
 // Libraries
 import { NativeModules, Platform } from 'react-native';
 import Package from 'react-native-package';
+import { extractProps } from './util/extractProps';
```
jmtaber129: Nit: vertical whitespace between third-party imports and first-party imports

js/util/__tests__/extractProps.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { Component, extractProps, PropExtractorConfig } from '../extractProps';
+import * as _ from 'lodash';
```
jmtaber129: Nit: separate third-party and first-party imports

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  moduleFileExtensions: ['ts', 'tsx', 'js'],
+  transform: {
+    '\\.(ts|tsx)$': 'ts-jest',
+  },
+  testRegex: '/__tests__/.*\\.(ts|tsx|js)$',
```
jmtaber129: I don't think this is consistent with our other repos, but it's fine for now. We should discuss folder architecture wrt tests (not super important, but probably easier to figure out now than migrate stuff later)

js/util/__tests__/extractProps.spec.ts
GrandPoohBear: Done.  I added an explicit test.

The `flatten` library handles this for us.  It'll just flatten an array like `a: [3,4,5]` into `{ "a.0": 3, "a.1": 4, "a.2": 5}`

js/util/extractProps.ts
GrandPoohBear: Just removed the TODO, but will keep that in mind.

js/util/extractProps.ts
GrandPoohBear: Sure, but I'm trying to be extra careful that this code won't crash a running app.  Typescript only gives you compile time checks - I still worry about the runtime.

js/util/extractProps.ts
GrandPoohBear: Flatten handles that for us, so I can add an explicit check for good measure, but we won't ever actually encounter it!

js/util/extractProps.ts
GrandPoohBear: Yeah, once we live in the prettier world, you either gotta be able to express it in a prettier config (which may be possible - not sure) or just get used to reading it :-)

js/Heap.js
GrandPoohBear: done

js/util/__tests__/extractProps.spec.ts
GrandPoohBear: done

jest.config.js
GrandPoohBear: Yeah, as long as this follows general open source conventions, I'm not as concerned with it being consistent with other internal Heap repos.  That's the default for jest, so I'd suggest we stick with it.

GrandPoohBear:
TBH, I'd rather keep the circle config in this PR.  My reasoning: this code contains tests, and I want them to be run every time I push.  I could make another branch that this one depends on etc. etc., but it's a lot of work for very little gain.

The circle config is all boilerplate, btw.  It's just generated by their new project generator.

js/util/__tests__/extractProps.spec.ts
```diff
@@ -71,6 +72,18 @@ describe('Extracting Props with a configuration', () => {
...
 
+  test('arrays flatten properly', () => {
+    const obj2 = _.merge({}, obj1, {
+      props: {
+        a: [3, 4, 5],
+      },
+    });
+
+    expect(extractProps('Element', obj2, config)).toEqual(
+      '[a.0=3];[a.1=4];[a.2=5];[c=true];'
```
jmtaber129: I'm not entirely sure this is actually correct.  Since heap doesn't really support array props, this is effectively meaningless unless index zero maps to one thing, index one to something else, etc.

This is fine for now, but we should discuss this.

js/util/extractProps.ts
jmtaber129: If there's a chance that `!component` could be `true`, the type should probably be changed to `Component | null` or `Component?`.

PR #21, opened 2019-02-04
Including prop extractor config from the class definition

GrandPoohBear:


js/util/extractProps.ts
```diff
@@ -22,11 +29,18 @@ export const extractProps = (
...
   config: PropExtractorConfig
 ): string => {
-  if (!component || !config[elementName]) {
+  if (!component) {
     return '';
   }
 
-  const filteredProps = pick(component.props, config[elementName].include);
+  let classCriteria: PropExtractorCriteria = { include: [] };
+  if (component.heapOptions) {
```
jmtaber129: Since `component.heapOptions` is of type `{ eventProps? }`, this should also assert that `component.heapOptions.eventProps` exists, otherwise `classCriteria` could be `undefined` instead of `{ include: [] }`.

js/util/extractProps.ts
```diff
@@ -22,11 +29,18 @@ export const extractProps = (
...
+    classCriteria = component.heapOptions.eventProps as PropExtractorCriteria;
+  }
+  let classConfig = { [elementName]: classCriteria };
```
jmtaber129: const

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
+import * as Collections from 'typescript-collections';
+import { PropExtractorCriteria, PropExtractorConfig } from './extractProps';
+
+export const getCombinedCriteria = (
+  componentName: string,
+  configs: PropExtractorConfig[]
+): PropExtractorCriteria => {
```
jmtaber129: Since this is really just producing a list of props to include (rather than a list to include and/or exclude) this should probably just return the inclusion array.

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+  let remainingConfigs: PropExtractorConfig[] = [...configs];
+  let inclusionSet = new Collections.Set<string>();
```
jmtaber129: The object is mutable, but the reference isn't, so this should be const.

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
+import * as Collections from 'typescript-collections';
+import { PropExtractorCriteria, PropExtractorConfig } from './extractProps';
+
+export const getCombinedCriteria = (
+  componentName: string,
+  configs: PropExtractorConfig[]
+): PropExtractorCriteria => {
+  let remainingConfigs: PropExtractorConfig[] = [...configs];
```
jmtaber129: const

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+  let inclusionSet = new Collections.Set<string>();
+
+  while (remainingConfigs.length > 0) {
+    let config = remainingConfigs.shift();
```
jmtaber129: const

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+    let criteria = config[componentName];
```
jmtaber129: const

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    if (!criteria) {
+      continue;
+    }
+
+    let configInclusionSet = new Collections.Set<string>();
```
jmtaber129: const

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+    let configExclusionSet = new Collections.Set<string>();
```
jmtaber129: const

js/util/extractProps.ts
```diff
@@ -6,11 +8,16 @@ export interface Component {
...
+}
+
+export interface ClassHeapOptions {
+  eventProps?: PropExtractorCriteria;
 }
 
 export interface PropExtractorCriteria {
   include: string[];
-  exclude: string[];
+  exclude?: string[];
```
jmtaber129: If `exclude` is optional, shouldn't `include` also be optional?  E.g., a higher priority config might only be concerned with excluding props, while a lower priority config might only be concerned with including props.

The way the include/exclude arrays should be represented should be one the following for both arrays:
* `undefined` means no props in that array
* empty array means no props in that array

When `include` is required, but `exclude` is optional, the way "no props" can be represented isn't consistent.

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+    criteria.include.map(s => configInclusionSet.add(s));
+    if (criteria.exclude) {
+      criteria.exclude.map(s => configExclusionSet.add(s));
+    }
+
+    inclusionSet.difference(configExclusionSet);
+    inclusionSet.union(configInclusionSet);
+  }
+
+  return { include: inclusionSet.toArray().sort() };
```
jmtaber129: Why does this need to be sorted?

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+    let config = remainingConfigs.shift();
+    let criteria = config[componentName];
+
+    if (!criteria) {
+      continue;
+    }
+
+    let configInclusionSet = new Collections.Set<string>();
+    let configExclusionSet = new Collections.Set<string>();
+    criteria.include.map(s => configInclusionSet.add(s));
```
jmtaber129: I don't think we have any specific guidance on this, but I generally prefer `forEach` when each element in an array is having some action performed with it (e.g. adding the element to a set), and `map` when each element in the array is being mapped to an element in a new array.

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
...
+import { PropExtractorCriteria, PropExtractorConfig } from './extractProps';
+
+export const getCombinedCriteria = (
+  componentName: string,
+  configs: PropExtractorConfig[]
+): PropExtractorCriteria => {
+  let remainingConfigs: PropExtractorConfig[] = [...configs];
+  let inclusionSet = new Collections.Set<string>();
+
+  while (remainingConfigs.length > 0) {
```
jmtaber129: Why not just `configs.forEach(config => {...})`?

js/util/combineCriteria.ts
```diff
@@ -0,0 +1,31 @@
+import * as Collections from 'typescript-collections';
+import { PropExtractorCriteria, PropExtractorConfig } from './extractProps';
+
+export const getCombinedCriteria = (
+  componentName: string,
+  configs: PropExtractorConfig[]
```
jmtaber129: Why not take an array of `PropExtractorCriteria`s instead?

js/util/extractProps.ts
GrandPoohBear: done - good catch.  I added a test to cover that case as well.

js/util/extractProps.ts
GrandPoohBear: done

js/util/combineCriteria.ts
GrandPoohBear: done - good catch

js/util/combineCriteria.ts
GrandPoohBear: Not as written, since it gets modified in the loop, but to be honest, I don't know why I took that approach, and it's kinda ugly.  Changing it to a for loop.

js/util/extractProps.ts
GrandPoohBear: No, I don't think so - I put the comment in the RFC on paper, but I'll copy it here in case you see this first:
No.  I think that explicit exclusions will not be often used, and so I’d like to avoid having the programmer need to write it each time.  So that should be optional. However, in the vast majority of cases, it’s pointless to do this without an inclusion set, and it will nearly always be there, so that should be required.  It may be empty, but there’s little gained by allowing it to be omitted, and we force any code that touches it to have to check that it’s defined.

js/util/combineCriteria.ts
GrandPoohBear: Since we're dealing with sets above, we don't get a consistent ordering.  This isn't strictly necessary but will make working with these things a little nicer in spots (testing, debug output, etc.) I could be convinced otherwise if you have a good reason.

js/util/combineCriteria.ts
GrandPoohBear: Fair - fixed.

js/util/combineCriteria.ts
GrandPoohBear: Could - and I think I had it that way at one point.  Not sure I can think of a good reason to change it, as long as the contract is solid.  It'd just be shifting the extraction logic around to other spots.

js/util/combineCriteria.ts
GrandPoohBear: Yeah - I like that.  Refactored with the last commit.

js/util/combineCriteria.ts
GrandPoohBear: Yeah, I can't honestly say...

GrandPoohBear: Fixed

js/util/combineCriteria.ts
jmtaber129: The reason I'm thinking is that `componentName` is only passed in to get the `PropExtractorCriteria` from the `PropExtractorConfig`.  Additionally, at the call site, `classCriteria` is getting converted from a `PropExtractorCriteria` to a `PropExtractorConfig` for the sole purpose of matching this function's signature.  So `classCriteria` is getting converted to a config, then back to a criteria.

To me, this method should just take an array of criteria, merge them together, and return the result.

js/util/combineCriteria.ts
jmtaber129: Not actually modified - it's mutated with `.shift()`, but its value is never reassigned.

js/util/combineConfigs.ts
```diff
@@ -0,0 +1,29 @@
...
+import { PropExtractorConfig } from './extractProps';
+
+export const getCombinedInclusionList = (
+  componentName: string,
+  configs: PropExtractorConfig[]
+): string[] => {
+  const inclusionSet = new Collections.Set<string>();
+
+  for (const config of configs) {
+    let criteria = config[componentName];
```
jmtaber129: const

js/util/extractProps.ts
jmtaber129: I still disagree.  Although we expect that criteria with only `include`s will be more common than only `exclude`s, this won't always be the case.  If we include builtin props that the dev needs to exclude, the criteria will likely only need an `exclude`.  It's inconsistent to make one of these required/optional but not both.  To me, consistency with this sort of thing is important when judging the quality of an API.

Adding code that checks that `include` is defined is cheap, especially if it's already checking whether `exclude` is defined.

js/util/combineCriteria.ts
jmtaber129: This just seems like YAGNI to me.  If we want ordering for testing/debug output/etc., the sorting should happen there, not here.

js/util/combineConfigs.ts
```diff
@@ -0,0 +1,29 @@
...
+    criteria.include.forEach(s => configInclusionSet.add(s));
+    if (criteria.exclude) {
+      criteria.exclude.forEach(s => configExclusionSet.add(s));
+    }
+
+    inclusionSet.difference(configExclusionSet);
+    inclusionSet.union(configInclusionSet);
+  }
+
+  return inclusionSet.toArray().sort();
```
drob: Could you leave a comment explaining the rationale for the sort?

js/util/combineConfigs.ts
GrandPoohBear: done

js/util/combineConfigs.ts
```diff
@@ -0,0 +1,29 @@
...
+  const inclusionSet = new Collections.Set<string>();
+
+  for (const config of configs) {
+    let criteria = config[componentName];
+
+    if (!criteria) {
+      continue;
+    }
+
+    let configInclusionSet = new Collections.Set<string>();
```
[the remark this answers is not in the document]
GrandPoohBear: done

js/util/combineConfigs.ts
```diff
@@ -0,0 +1,29 @@
...
+    let configExclusionSet = new Collections.Set<string>();
```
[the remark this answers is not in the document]
GrandPoohBear: done

js/util/combineConfigs.ts
GrandPoohBear: done

js/util/combineCriteria.ts
GrandPoohBear: done

js/util/combineCriteria.ts
GrandPoohBear: Added a comment.

js/util/extractProps.ts
GrandPoohBear: Left as-is per discussion.

js/util/__tests__/combineConfigs.spec.ts
```diff
@@ -1,30 +1,48 @@
...
   test('combines configs normally', () => {
-    expect(
-      getCombinedInclusionList('FooClass', [config1, config2, config3])
-    ).toEqual(['c', 'e', 'f', 'g']);
+    expect(getCombinedInclusionList([criteria1, criteria2, criteria3])).toEqual(
+      ['c', 'e', 'f', 'g']
+    );
+  });
+
+  test("Can't include something previously excluded", () => {
```
jmtaber129: Nit: "Can't include" -> "doesn't include"

js/util/__tests__/combineConfigs.spec.ts
```diff
@@ -1,30 +1,48 @@
...
 
-const makeTestConfig = (include, exclude) => {
+const makeTestCriteria = (include, exclude) => {
   return deepFreeze({
-    FooClass: {
-      include: include,
-      exclude: exclude,
-    },
+    include: include,
+    exclude: exclude,
```
jmtaber129: Optional nit: I think this can be written as just
```
{
  include,
  exclude,
}
```

PR #23, opened 2019-02-07
Ignore 'View' components in hierarchy

jmtaber129:
`View` components show up in the hierarchy in dev builds, but not prod builds, and aren't actually relevant to the overall component hierarchy.  Leave `View`s out of the hierarchy.

### DIFF
--- a/js/Heap.js
+++ b/js/Heap.js
@@ -99,6 +99,11 @@ const getFiberNodeComponentHierarchy = currNode => {
 
   const elementName =
     currNode.elementType.displayName || currNode.elementType.name;
+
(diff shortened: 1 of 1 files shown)


js/Heap.js
```diff
@@ -99,6 +99,11 @@ const getFiberNodeComponentHierarchy = currNode => {
 
   const elementName =
     currNode.elementType.displayName || currNode.elementType.name;
+
+  if (elementName === 'View') {
```
drob: Could we get a comment explaining this special case?

drob:
LGTM.

PR #24, opened 2019-02-07
Capture target text on autotrack events

jmtaber129:
Determine target text by concatenating target texts of component children, and include it in the autotracked event.

### DIFF
--- a/js/Heap.js
+++ b/js/Heap.js
@@ -62,14 +62,41 @@ export default Package.create({
           componentThis.state.touchable &&
           componentThis.state.touchable.touchState;
 
+        const targetText = getTargetText(componentThis._reactInternalFiber);
+
         track(eventType, {
           touchableHierarchy,
           touchState,
+          targetText,
         });
       },
     };
   },
 });
+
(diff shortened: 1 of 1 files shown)


js/Heap.js
```diff
@@ -62,15 +62,42 @@ export default Package.create({
...
           touchState,
+          targetText,
         });
       },
     };
   },
 });
 
+// :TODO: (jmtaber129): Consider implementing sibling target text.
+const getTargetText = fiberNode => {
```
drob: If the target text is empty, we're returning `''` which will be a little strange downstream in our UI. I propose we instead set it to null and filter out the prop from the blob entirely, in the `track` call. Wdyt?

drob: Oh, I see – you're using the empty but non-null string for purposes of recursion. We could do something outside of here in the caller to not send `''` to the server. Thoughts?

jmtaber129: Yea, when there's no target text on the web, that prop is just not sent.  Updated to only include that prop when it's non-empty

drob:
LGTM

PR #25, opened 2019-02-08
Get component name from 'type' instead of 'elementType'

jmtaber129:
It seems like `elementType` isn't present on fiberNodes for certain versions and/or configs, but `type` does seem to be present regardless of whether `elementType` is present.


drob:
LGTM.

PR #26, opened 2019-02-11
Remove react-native-package dependency

jmtaber129:
The main reason this dependency was originally included was to restrict the platforms where rnheap would run to only iOS (since android wasn't implemented).  Now that android is also supported, we can remove this dependency.


drob:
LGTM.

PR #31, opened 2019-02-14
DO NOT MERGE: POC React Navigation autotrack

jmtaber129:
This includes a POC for autotrack of screenviews for apps using `react-navigation`.  It uses the `onNavigationStateChange` of `react-navigation`'s `NavigationContainer` (created by `createAppContainer`) to listen for navigation events that look like screenview transitions, and then determine the path in the navigation router, of the format `<top route name>::<route name>::<route name>::<etc.>`.  For example, when a navigation event leads to a screenview on the `MainScreen` in the `StackNav` navigator within the `AppNavigator` within the `TopNavigator`, the path would be `Main::Main::Original`, since `Main` is the name of the `AppNavigator` route, `Main` is the name of the `StackNav` route, and `Original` is the name of the `MainScreen` route.

This listener is added by creating a HOC of the app's `NavigationContainer` using a method currently named `withHeapNavigationAutotrack()` (note that this doesn't currently correctly pass through props/etc. like a HOC wrapper should).

The two relevant files are `examples/TestDriver/App.js` (where routing + additional screens were added) and `js/Heap.js` (where the HOC wrapper function + tracking utils were added).

### DIFF
--- a/examples/TestDriver/android/app/src/main/java/com/testdriver/MainApplication.java
+++ b/examples/TestDriver/android/app/src/main/java/com/testdriver/MainApplication.java
@@ -3,6 +3,8 @@ package com.testdriver;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.swmansion.gesturehandler.react.RNGestureHandlerPackage;
+import com.heapanalytics.reactnative.RNHeapLibraryPackage;
 import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
 import com.facebook.react.shell.MainReactPackage;
@@ -24,6 +26,8 @@ public class MainApplication extends Application implements ReactApplication {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
               new MainReactPackage(),
+            new RNGestureHandlerPackage(),
+            new RNHeapLibraryPackage(),
               new RNHeapLibraryPackage()
       );
     }
--- a/examples/TestDriver/android/settings.gradle
+++ b/examples/TestDriver/android/settings.gradle
@@ -1,4 +1,8 @@
 rootProject.name = 'TestDriver'
+include ':react-native-gesture-handler'
+project(':react-native-gesture-handler').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-gesture-handler/android')
+include ':@heap_react-native-heap'
+project(':@heap_react-native-heap').projectDir = new File(rootProject.projectDir, '../node_modules/@heap/react-native-heap/android')
 
 include ':app'
 
--- a/examples/TestDriver/ios/TestDriver/AppDelegate.m
+++ b/examples/TestDriver/ios/TestDriver/AppDelegate.m
@@ -33,10 +33,10 @@
   [self.window makeKeyAndVisible];
 
   // Send events to local collector.
-  SEL setRootUrlSelector = @selector(setRootUrl:);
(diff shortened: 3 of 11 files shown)


examples/TestDriver/android/app/src/main/java/com/testdriver/MainApplication.java
```diff
@@ -24,6 +26,8 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
               new MainReactPackage(),
+            new RNGestureHandlerPackage(),
+            new RNHeapLibraryPackage(),
```
GrandPoohBear: Also a duplicate of the line immediately following, and should also be removed.

examples/TestDriver/android/settings.gradle
```diff
@@ -1,4 +1,8 @@
 rootProject.name = 'TestDriver'
+include ':react-native-gesture-handler'
+project(':react-native-gesture-handler').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-gesture-handler/android')
+include ':@heap_react-native-heap'
```
GrandPoohBear: This should just be fixed on line 8-10 below.

examples/TestDriver/ios/TestDriver/AppDelegate.m
```diff
@@ -33,10 +33,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   [self.window makeKeyAndVisible];
 
   // Send events to local collector.
-  SEL setRootUrlSelector = @selector(setRootUrl:);
-  if ([[Heap class] respondsToSelector:setRootUrlSelector]) {
-    [[Heap class] performSelector:setRootUrlSelector withObject:@"http://localhost:3000"];
-  }
+//  SEL setRootUrlSelector = @selector(setRootUrl:);
```
GrandPoohBear: Maybe we should just do this with a top-level boolean so that we can just turn it on or off with a single line...

js/Heap.js
```diff
@@ -12,6 +13,60 @@ const track = (event, payload) => {
...
+}
+
+let navRef = null;
+const NavigationService = {
+  setTopLevelNavigator: (ref) => {
+    navRef = ref;
+  }
+}
+
+const withHeapNavigationAutotrack = (AppContainer) => {
```
GrandPoohBear: I'd love to see this move to its own file so we can keep this one succinct (and the file containing the nav stuff separate).

GrandPoohBear: Also, since this is going to be specific to React Navigation, we should try to put that in the name of the thing somehow (as ungainly as that's going to be).  Maybe `withReactNavigationAutotrack` (I suspect we can drop the "Heap" since that will be obvious from the way it's used, i.e. Heap.withReactNavigationAutotrack)?

js/Heap.js
```diff
@@ -12,6 +13,60 @@ const track = (event, payload) => {
...
+        path: initialPageviewPath,
+      });
+      console.log(initialPageviewPath);
+    }
+
+    render() {
+      return (
+        <AppContainer
+          ref={navigatorRef => {
+            NavigationService.setTopLevelNavigator(navigatorRef);
```
GrandPoohBear: This is interesting.  Why do we need to do this?  Would it interact with any other configuration in the app?

GrandPoohBear:
Closing - this is productized in https://github.com/heap/react-native-heap/pull/46

PR #33, opened 2019-02-15
Move detox tests from internal repo

jmtaber129:
To make it easier to keep e2e tests in-sync with the library, we're moving the detox tests from the internal repo to the public repo.

Because the test assertions depend on internal utils, the tests can only be run by those with access to the internal repo (i.e. Heap engineers).

Additional details are in the corresponding PR in the internal repo.

### DIFF
--- a/e2e/firstTest.spec.js
+++ b/e2e/firstTest.spec.js
@@ -1,0 +1,240 @@
+require('coffeescript').register();
+_ = require('lodash');
+assert = require('should/as-function')
+
+db = require('../../heap/back/db');
+testUtil = require('../../heap/test/util');
+
+const assertEvent = (err, res, check) => {
+  assert.not.exist(err);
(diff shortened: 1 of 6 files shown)


e2e/firstTest.spec.js
```diff
@@ -0,0 +1,240 @@
+require('coffeescript').register();
```
GrandPoohBear: Is the coffeescript necessary for the utility libraries out of heap/heap?

jmtaber129: Yep.  The utilities are in coffeescript, and detox doesn't recognize coffeescript files without `coffeescript/register`

jmtaber129:
Yea, a checkbox for non-CI tests probably would be a good thing.  Let's chat about this at the next daily.

PR #36, opened 2019-02-20
Add directive to force RNHeap to be built as a static framework

GrandPoohBear:


examples/TestDriver/ios/Podfile
```diff
@@ -35,3 +35,11 @@ target "TestDriver" do
...
   end
 end
+
+# Force react-native-heap to be built as a static framework
+pre_install do |installer|
+    pod = installer.pod_targets.find { |p| p.name == 'react-native-heap'}
+    def pod.static_framework?
+        true
+    end
+end
```
jmtaber129: If we're adding this, should we add `use_frameworks!` at the top, too?

README.md
```diff
@@ -114,10 +114,34 @@ Heap.track('signed-up', { isPaid: true, amount: 20 });
...
+  "_OBJC_CLASS_$_Heap", referenced from:
+      objc-class-ref in RNHeap.o
+ld: symbol(s) not found for architecture x86_64
+clang: error: linker command failed with exit code 1 (use -v to see invocation)
+```
+
+This may appear if the `Podfile` contains a `use_frameworks!` directive. One solution is to add the following at the very end of your `Podfile`:
+
+```ruby
+# Force react-native-heap to be built as a static framework
```
jmtaber129: Nit: period at end of comment.

examples/TestDriver/ios/Podfile
GrandPoohBear: Added.  It shouldn't matter to much except as an example, so might as well make it as close to what a user might encounter as possible.

PR #37, opened 2019-02-21
Allow prop capture configs to be attached to functional components

jmtaber129:
Currently, there is no way to attach prop capture configs to functional components, since we look for `heapOptions` on `stateNode`s, which do not exist on functional component instances.

With a functional component like
```
const Foo = (props) => {
  return (
    <some jsx>
  );
};
```
a `heapOptions` object can now be added by doing
```
Foo.heapOptions = {
   <options>
}
```
This would lead to `heapOptions` showing up on the `fiberNode.type` object.

While it's technically possible to allow this for class components, doing so could be potentially error-prone, since it would provide two different ways to declare heap options that wouldn't be close to each other in code.  E.g. a dev might add heap options as a static prop on a class component when heap options already exist through `Foo.heapOptions = {...}` without even realizing it.

Note: Because of the `npm start` issue with `TestDriver` on my machine, e2e tests have **not** been run.  These should still be run before merging (although I don't expect any issues).

### DIFF
--- a/js/util/extractProps.ts
+++ b/js/util/extractProps.ts
@@ -17,7 +17,7 @@ export interface FiberNode {
   stateNode?: StateNode;
   type?: {
     heapOptions?: ClassHeapOptions;
-  }
+  };
 }
 
 export interface ClassHeapOptions {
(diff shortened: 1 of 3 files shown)


js/util/extractProps.ts
```diff
@@ -42,8 +45,21 @@ export const extractProps = (
   }
 
   let classCriteria: PropExtractorCriteria = EMPTY_CRITERIA;
-  if (fiberNode.stateNode && fiberNode.stateNode.heapOptions && fiberNode.stateNode.heapOptions.eventProps) {
-    classCriteria = fiberNode.stateNode.heapOptions.eventProps as PropExtractorCriteria;
+  if (
```
GrandPoohBear: I think it would be good to add a comment here around why we are expecting these things to exist, and on which components (e.g. Class based components will have a fiberNode, and functional components will not, etc.)

jmtaber129: done

PR #38, opened 2019-02-26
[HEAP-8030] Allowing either specifying an entire nested object, or by flattened key.

GrandPoohBear:
@jnatkins raised the point that we might want to allow users to specify sub-attributes of nested components piecemeal, instead of specifying that we want the entire nested object.  This requires a two-pass approach on grabbing the props (since we didn't actually want to lose the other behavior either).  The first pass grabs the whole object, and then after the input object is flattened, we make another pass and combine the two for the end result.

### DIFF
--- a/js/util/__tests__/extractProps.spec.ts
+++ b/js/util/__tests__/extractProps.spec.ts
@@ -24,6 +24,25 @@ describe('Extracting Props with a configuration', () => {
     },
   };
 
+  const objWithNestedObject = _.merge({}, obj1, {
+    stateNode: {
+      props: {
+        a: {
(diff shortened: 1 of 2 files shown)


GrandPoohBear:
Oh, interesting!  Yeah, TBH, I don't care _that_ much about keeping our dependencies small.  It's generally less of an issue for mobile folks than web.  (I also know a guy who's a bit of a zealot about these things, so I might pick his brain on what he would want to see)

GrandPoohBear:
Closing, in favor of https://github.com/heap/react-native-heap/pull/39

PR #43, opened 2019-03-05
Add detox tests for property extraction

GrandPoohBear:
This PR also includes a number of changes along the way, and in general, they're split out by commit for cleanliness.  There are:

- Purely formatting changes (commits with `prettier` in the title)
- Addition of React Navigation, so that we could keep test pages separate
- Slight refactors to existing tests, to pull out shared logic into common module

The detox tests now run two "pages" of tests: the first one named "Basics" holds the previous code, and includes basic configuration and touchable autotracking.  The "PropExtraction" tab is new, and covers tests for configurable and automatic property extraction in the hierarchy.  Unfortunately, since the "Button" component is slightly different between platforms out of the box, there is a small bit of logic that differs between platforms, but otherwise the tests are very straightforward.  Detox simulates a tap on a button, we wait until we're sure the events have been sent, and then match exactly what we're expecting for the component hierarchy in the event.

### DIFF
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -6,6 +6,7 @@ testUtil = require('../../heap/test/util');
 rnTestUtil = require('./rnTestUtilities');
 
 const doTestActions = async () => {
+  // Open the Basics tab in the tab navigator.
   await element(by.id('Basics')).tap();
 
   await expect(element(by.id('track1'))).toBeVisible();
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -6,6 +6,8 @@ testUtil = require('../../heap/test/util');
 rnTestUtil = require('./rnTestUtilities');
 
 const doTestActions = async () => {
+  await element(by.id('Basics')).tap();
+
   await expect(element(by.id('track1'))).toBeVisible();
   await element(by.id('track1')).tap();
   await element(by.id('aep')).tap();
@@ -37,11 +39,7 @@ describe('Basic React Native and Touchable Support', () => {
   describe(':ios: Bridge API', () => {
     before(async () => {
       await doTestActions();
-      // Heap iOS flushes events every 15 seconds. Wait 16 seconds to ensure that
-      // all events are flushed to redis before asserting.
-      // :TODO:(jmtaber129): Make this wait shorter if/when we expose setting the
-      // flush frequency to the RN bridge.
-      await new Promise(resolve => setTimeout(resolve, 16000));
+      await rnTestUtil.waitForEventsToFlush();
     });
 
     it('should call first track', async () => {
@@ -121,15 +119,11 @@ describe('Basic React Native and Touchable Support', () => {
   describe(':android: Bridge API', () => {
     before(async () => {
       await doTestActions();
-      // Heap Android flushes events every 15 seconds. Wait 16 seconds to ensure that
-      // all events are flushed to redis before asserting.
-      // :TODO:(jmtaber129): Make this wait shorter if/when we expose setting the
-      // flush frequency to the RN bridge.
-      await new Promise(resolve => setTimeout(resolve, 16000));
+      await rnTestUtil.waitForEventsToFlush();
     });
 
     it('should call first track', async () => {
-      await assertAndroidEvent(
+      await rnTestUtil.assertAndroidEvent(
         {
           envId: '2084764307',
           event: { custom: { name: 'pressInTestEvent1' } },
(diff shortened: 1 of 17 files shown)


examples/TestDriver/src/topNavigator.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { createBottomTabNavigator, createAppContainer } from 'react-navigation';
+
+import TouchablesPage from './pages/touchables';
+
+import Heap from '@heap/react-native-heap';
+
+Heap.setAppId('2084764307');
```
drob: Why are we setting this statically, vs in a `componentDidMount` or something?

I don't doubt there's a good reason but can you walk me through it?

e2e/firstTest.spec.js
```diff
@@ -298,7 +298,7 @@ describe('ReactNative Support', () => {
   describe('Autotrack', () => {
     it("should autotrack 'TouchableOpacity's", async () => {
       const expectedHierarchy =
-        'AppContainer;|App;|Provider;|Connect(MainScreen);|MainScreen;|TouchableOpacity;|';
+        'AppContainer;|App;|Provider;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;|SceneView;|Connect(TouchablesPage);|TouchablesPage;|TouchableOpacity;|';
```
drob: Wow that's a lot of added hierarchy for wrapping this app in a tab navigator! O_O

Wonder how soon we're going to have to start adding easy ways to filter out noise in these / simplify them? I suppose the length doesn't matter if they're stable and you're using the right tools to create the defs.

e2e/basics.spec.js
```diff
@@ -6,6 +6,8 @@ testUtil = require('../../heap/test/util');
 rnTestUtil = require('./rnTestUtilities');
 
 const doTestActions = async () => {
+  await element(by.id('Basics')).tap();
```
drob: This is to open the `Basics` tab but doesn't do any events we expect to record below?

Assuming so, might be helpful to comment to that effect.

Also, assuming so, will this start triggering something once react navigation capture is merged?

examples/TestDriver/src/pages/propExtraction.js
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
```
drob: Is there an exclude case that would be worth testing somewhere in here?

examples/TestDriver/src/topNavigator.js
GrandPoohBear: No particular reason, other than adding a React Navigation "wrapper" on things means there isn't really a convenient `componentDidMount` to put it in.  This will essentially have the same effect, and should only be run once (but is also idempotent, so doesn't really matter.)

e2e/firstTest.spec.js
GrandPoohBear: Yeah, that is something I still don't have the best understanding of.  This seems similar to CSS hierarchies in terms of complexity, no?

e2e/basics.spec.js
GrandPoohBear: Comments added.

Yeah, these will likely get autotracked along with any of the other ReactNavigation events.

examples/TestDriver/src/pages/propExtraction.js
GrandPoohBear: Good call - added a test for that

e2e/propExtraction.spec.js
```diff
@@ -63,5 +64,20 @@ describe('Property Extraction in Hierarchies', () => {
         expectedTargetText
       );
     });
+
+    it('properly excludes properties', async () => {
+      // The important thing for this test is that the first mention of 'Button' does NOT include the title property.
+      // This is because the first one is a custom class that specifically excludes (while the second one is the built-in Button.)
```
drob: I was confused until I read this. This makes sense. 👍

e2e/firstTest.spec.js
drob: Lots of CSS hierarchies are much worse for sure. I think this was the first time I saw a real tab nav hierarchy so the length surprised me.

e2e/propExtraction.spec.js
GrandPoohBear: Yeah, since currently the title on Button is the _only_ thing that's possible to exclude (and actually see a difference!)

PR #50, opened 2019-03-12
Make sure we don't barf on an empty prop payload to track

GrandPoohBear:


js/__tests__/Heap.spec.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { NativeModules } from 'react-native';
+
+import Heap from '../Heap';
+
+describe('The Heap object', () => {
```
afrieder: It'd be nice if there were some tests for the happy path with more than an empty object.

GrandPoohBear: Oh, duh... nice catch... will add and merge

PR #51, opened 2019-03-12
Use a shorter timer in native apps so events flush quicker for tests

GrandPoohBear:


examples/TestDriver/android/app/build.gradle
```diff
@@ -146,6 +146,7 @@ android {
 }
 
 dependencies {
+    implementation 'com.heapanalytics.android:heap-android-client:0.8.3'
```
apyousef: You don't want to add this here. You want to modify `react-native-heap/android/build.gradle` to have the following line:
`api 'com.heapanalytics.android:heap-android-client:0.8.3'`

This would replace the `implementation... line`.

See [here](https://developer.android.com/studio/build/dependencies#dependency_configurations). Try that out locally and test it to see if it does what you want.

drob:
I can't speak to whether or not this is the right way to do this, but 🙌that we're doing this!!

PR #52, opened 2019-03-12
Add detox e2e tests for React Navigation

GrandPoohBear:
The first commit refactors the existing tests slightly, renaming some of the files and moving the navigationOptions into a common spot in the navigator.  The second commit adds the tests for navigation, and the pages to support it.  To capture the initial navigation event, there is a new "initial" page with just a single test.

### DIFF
--- a/e2e/00_initial.spec.js
+++ b/e2e/00_initial.spec.js
@@ -1,0 +1,24 @@
+require('coffeescript').register();
+_ = require('lodash');
+assert = require('should/as-function');
+
+testUtil = require('../../heap/test/util');
+rnTestUtil = require('./rnTestUtilities');
+
+// This file is named funkily so that Mocha runs this first.  That way we
+// don't lose the initial navigation event (each test flushes redis).
+
+describe('Initial Navigation', () => {
+  before(async () => {
+    await device.launchApp();
+
+    await expect(element(by.id('initialSentinel'))).toBeVisible();
+    await element(by.id('initialSentinel')).tap();
+
+    await rnTestUtil.pollForSentinel('Initial');
+  });
+
+  it('tracks the initial route', async () => {
+    await rnTestUtil.assertNavigationEvent('Initial');
+  });
+});
--- a/e2e/00_initial.spec.js
+++ b/e2e/00_initial.spec.js
@@ -1,6 +1,4 @@
 require('coffeescript').register();
-_ = require('lodash');
-assert = require('should/as-function');
 
 testUtil = require('../../heap/test/util');
 rnTestUtil = require('./rnTestUtilities');
--- a/e2e/nav.spec.js
+++ b/e2e/nav.spec.js
@@ -46,6 +46,7 @@ describe('Navigation', () => {
         'Navigation/NAVIGATE'
       );
     });
+
     it('tracks a stack navigator action', async () => {
       await rnTestUtil.assertNavigationEvent(
         'Nav::MainStack::StackCard',
--- a/e2e/nav.spec.js
+++ b/e2e/nav.spec.js
@@ -1,0 +1,68 @@
+require('coffeescript').register();
+_ = require('lodash');
+assert = require('should/as-function');
+
+testUtil = require('../../heap/test/util');
+rnTestUtil = require('./rnTestUtilities');
+
+const delay = async () => {
+  await new Promise(resolve => setTimeout(resolve, 500));
+};
+
+const doTestActions = async () => {
+  // Open the PropExtraction tab in the tab navigator.
+  await element(by.id('Nav')).tap();
+
+  await expect(element(by.id('navigate_stack'))).toBeVisible();
+  await element(by.id('navigate_stack')).tap();
+  await delay();
+  await expect(element(by.id('pop1'))).toBeVisible();
+  await element(by.id('pop1')).tap();
+  await delay();
+  await expect(element(by.id('navigate_modal'))).toBeVisible();
+  await element(by.id('navigate_modal')).tap();
+  await delay();
+  await expect(element(by.id('pop2'))).toBeVisible();
+  await element(by.id('pop2')).tap();
+  await delay();
+
+  await element(by.id('navSentinel')).tap();
+};
+
+describe('Navigation', () => {
+  before(done => {
+    db.orm.connection.sharedRedis().flushall(done);
+  });
+
+  describe('React Navigation autotrack', () => {
+    before(async () => {
+      await doTestActions();
+      await rnTestUtil.pollForSentinel('Nav');
+    });
+
+    it('tracks a tab navigator action', async () => {
+      await rnTestUtil.assertNavigationEvent(
+        'Nav::MainStack::Base',
+        'Navigation/NAVIGATE'
+      );
+    });
+    it('tracks a stack navigator action', async () => {
+      await rnTestUtil.assertNavigationEvent(
+        'Nav::MainStack::StackCard',
+        'Navigation/NAVIGATE'
+      );
+
+      await rnTestUtil.assertNavigationEvent(
+        'Nav::MainStack::Base',
+        'Navigation/POP'
+      );
+    });
+
+    it('tracks a modal stack navigator action', async () => {
+      await rnTestUtil.assertNavigationEvent(
+        'Nav::ModalStack',
+        'Navigation/NAVIGATE'
+      );
+    });
+  });
+});
--- a/e2e/rnTestUtilities.js
+++ b/e2e/rnTestUtilities.js
@@ -80,12 +80,116 @@ const assertAutotrackHierarchy = async (
   }
 };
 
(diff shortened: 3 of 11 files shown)


e2e/rnTestUtilities.js
```diff
@@ -80,19 +80,125 @@ const assertAutotrackHierarchy = async (
...
+  } else {
+    throw new Error(`Unknown device type: ${device.getPlatform()}`);
+  }
+};
+
 waitForEventsToFlush = async () => {
-  // Heap for iOS and Android flushes events every 15 seconds. Wait 16 seconds to ensure that
-  // all events are flushed to redis before asserting.
-  // :TODO:(jmtaber129): Make this wait shorter if/when we expose setting the
-  // flush frequency to the RN bridge.
```
drob: Given that we're removing this TODO, should we be lowering the delay to ~2 seconds?

Also nit re the replacement comment – not sure if there's a second half of that sentence.

drob: Discussed in person – this is probably moot now.

e2e/nav.spec.js
```diff
@@ -0,0 +1,68 @@
...
+      await rnTestUtil.pollForSentinel('Nav');
+    });
+
+    it('tracks a tab navigator action', async () => {
+      await rnTestUtil.assertNavigationEvent(
+        'Nav::MainStack::Base',
+        'Navigation/NAVIGATE'
+      );
+    });
+    it('tracks a stack navigator action', async () => {
```
awhitty: Lil nit but probably want newline above this line for consistency.

e2e/rnTestUtilities.js
```diff
@@ -80,12 +80,111 @@ const assertAutotrackHierarchy = async (
...
 };
 
-waitForEventsToFlush = async () => {
-  // Heap for iOS and Android flushes events every 15 seconds. Wait 16 seconds to ensure that
-  // all events are flushed to redis before asserting.
-  // :TODO:(jmtaber129): Make this wait shorter if/when we expose setting the
-  // flush frequency to the RN bridge.
-  await new Promise(resolve => setTimeout(resolve, 16000));
+const assertAndroidNavigationEvent = async (expectedPath, expectedType) => {
+  let props = {
```
awhitty: Style nit: could get away with a `const` by extracting the initial `{ path: ... }` object to another `const commonProps` or similar.

e2e/rnTestUtilities.js
```diff
@@ -80,12 +80,111 @@ const assertAutotrackHierarchy = async (
...
+      custom: {
+        name: 'reactNavigationScreenview',
+        properties: props,
+      },
+    },
+  });
+};
+
+const assertIosNavigationEvent = async (expectedPath, expectedType) => {
+  let props = ['path', expectedPath];
```
awhitty: Similar opportunity for `const` here.

e2e/nav.spec.js
```diff
@@ -0,0 +1,68 @@
...
+  await expect(element(by.id('pop2'))).toBeVisible();
+  await element(by.id('pop2')).tap();
+  await delay();
+
+  await element(by.id('navSentinel')).tap();
+};
+
+describe('Navigation', () => {
+  before(done => {
+    db.orm.connection.sharedRedis().flushall(done);
```
awhitty: 🤔 how is `db` in scope here?

e2e/rnTestUtilities.js
```diff
@@ -80,12 +80,111 @@ const assertAutotrackHierarchy = async (
...
+  const startTick = Date.now();
+
+  // Give it a few seconds at first.
+  await new Promise(resolve => setTimeout(resolve, 3000));
+
+  while (Date.now() - startTick <= timeout) {
+    const eventName = `${sentinelValue.toUpperCase()}_SENTINEL`;
+    if (device.getPlatform() === 'ios') {
+      const event = {
+        a: '2084764307',
```
awhitty: Seems unrelated to PR, but could probably extract this to some ALL_CAPS const and reuse it.

examples/TestDriver/src/pages/nav.js
```diff
@@ -0,0 +1,69 @@
...
+        title="Navigate Stack"
+        testID="navigate_stack"
+        onPress={() => props.navigation.navigate('StackCard')}
+      />
+      <Button
+        title="Navigate Modal"
+        testID="navigate_modal"
+        onPress={() => props.navigation.navigate('ModalStack')}
+      />
+      {makeSentinelButton('Nav')}
```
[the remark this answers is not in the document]
awhitty: Ah I see they're tested by expecting navigation events. Something feels a little fishy about how this `makeSentinelButton` thing is different than the other things. What exactly is it meant to do?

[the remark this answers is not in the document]
awhitty: Ah I see - I gather it's meant to signal that all the test actions have been completed. This makes sense to me now.

e2e/nav.spec.js
GrandPoohBear: detox magic!

e2e/nav.spec.js
```diff
@@ -0,0 +1,68 @@
+require('coffeescript').register();
+_ = require('lodash');
```
awhitty: Also unused I think?

e2e/00_initial.spec.js
```diff
@@ -0,0 +1,24 @@
+require('coffeescript').register();
+_ = require('lodash');
```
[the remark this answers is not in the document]
GrandPoohBear: I think you're right... it was just copy/pasta'd over here from somewhere else that needed it!

e2e/00_initial.spec.js
```diff
@@ -0,0 +1,24 @@
...
+assert = require('should/as-function');
```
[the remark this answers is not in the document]
GrandPoohBear: No, I think it's just for the assert, which I'm not using here.

e2e/rnTestUtilities.js
```diff
@@ -80,12 +80,111 @@ const assertAutotrackHierarchy = async (
...
+};
+
+pollForSentinel = async (sentinelValue, timeout = 60000) => {
+  console.log(
+    `--- Waiting for ${sentinelValue} sentinel event.  This will timeout in 60s ---`
+  );
+  const startTick = Date.now();
+
+  // Give it a few seconds at first.
+  await new Promise(resolve => setTimeout(resolve, 3000));
```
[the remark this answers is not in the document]
GrandPoohBear: There will always be _some_ amount of delay between generating the events on the phone simulator and getting them to be readable.  In my testing, it was closer to somewhere between 5-10s.  It's not strictly necessary, but not harmful either.

examples/TestDriver/src/pages/nav.js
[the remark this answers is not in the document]
GrandPoohBear: Yeah.  So previously we were just waiting for 16s (the default is to flush events every 15s) and then hoping they'd be there.  The sentinel event is the _last_ event that gets sent on any page, and signals that all of the previous ones should now be available.  The test polls for that event instead of just waiting, and then moves ahead when it sees it.  This cuts the wait times significantly.

e2e/nav.spec.js
GrandPoohBear: Oh yeah!  Thanks.  It's actually funny - I've got a jest runner plugin for VSCode that automatically inserts one-off "debug THIS test" buttons whenever it sees test-like stuff in code.  I've attached a screenshot of why I didn't think to put a newline here :-P
Screen Shot 2019-03-14 at 4 40 30 PM
So that "Debug" button _looks_ like a newline, but isn't!

e2e/rnTestUtilities.js
awhitty: In more detail, here's the sort of change I'd suggest, but it's mostly a style preference. I don't have a strong opinion here.

```ts
const commonProps = {
  path: {
    string: expectedPath,
  }
};

const props = expectedType
	? commonProps
	: { 
      ...commonProps, 
      type: {
        string: expectedType,
      }
    };
```

Could also just go for a little duplication to reduce cognitive load of what's goin on in that spread syntax.

```ts
const props = expectedType
  ? {
      path: {
        string: expectedPath
      }
    }
  : {
      path: {
        string: expectedPath
      },
      type: {
        string: expectedType
      }
    };
```

GrandPoohBear: done!

e2e/rnTestUtilities.js
GrandPoohBear: done

e2e/rnTestUtilities.js
GrandPoohBear: Yeah, totally... done...

examples/TestDriver/src/pages/nav.js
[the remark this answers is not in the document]
awhitty: Got it! Yup definitely sounds like an improvement.

PR #56, opened 2019-03-14
Fix an issue with addUserProperties and addEventProperties

GrandPoohBear:


js/Heap.js
```diff
@@ -5,11 +5,12 @@ import { NativeModules } from 'react-native';
 import { autotrackPress } from './autotrack/touchables';
 import { withReactNavigationAutotrack } from './autotrack/reactNavigation';
 
+const flatten = require('flat');
 const RNHeap = NativeModules.RNHeap;
 
 const track = (event, payload) => {
   try {
-    let flatten = require('flat');
+    const flatten = require('flat');
```
awhitty: Is this require still needed?

GrandPoohBear: Yeah, and I was just thinking I should write a comment.  Doing it this way allows me to mock the flatten function in a test where I make that mysteriously fail, just to demonstrate that the entire process doesn't crash - just logs an error.

PR #60, opened 2019-03-18
Warning for missing Display Names

GrandPoohBear:


js/util/checkDisplayNames.ts
```diff
@@ -0,0 +1,15 @@
...
+  render() {}
+}
+
+let warningGiven = false;
+
+export const checkDisplayNamePlugin = () => {
+  // @ts-ignore
+  if (!DisplayNameTest.displayName && !warningGiven) {
+    console.warn(
+      'Heap: Display names are not being generated for React components.  Make sure the "add-react-displayname" plugin is specified in your babel config, or React hierarchies will have dummy component names.'
```
jnatkins: Recommend slight rewording:

Heap: Display names are not being generated for React components, which will cause Heap to populate captured component hierarchies with dummy component names.  Make sure the "add-react-displayname" plugin is specified in your babel config, and that the config is being properly picked up by the app.

PR #63, opened 2019-03-29
Version bump and docs change for 0.2.0

GrandPoohBear:


README.md
```diff
@@ -43,6 +41,19 @@ pod install
 
 You're done! :tada:
 
+### Manual setup - iOS without Cocoapods
+
+All terminal commands assume you are in the top-level directory of your React Native project.
+
+1. Locate the Xcode project for RNHeap: `open node_modules/@heap/react-native-heap/ios` . You should see a Finder window containing the `RNHeap.xcodeproj` file.
+1. Open the Xcode project for your native app, in the `ios` directory of your project.
```
jnatkins: This might be a markdown thing that I'm just not positive on, but these will show up as incremental numbers, right?

GrandPoohBear: Yup!  It's actually a nice trick - that way you can insert one in the middle, and all of the numbers will update properly.

PR #64, opened 2019-03-29
Upgrade detox and add delays to make tests pass

GrandPoohBear:


e2e/basics.spec.js
```diff
@@ -13,12 +13,22 @@ const doTestActions = async () => {
   await element(by.id('track1')).tap();
   await element(by.id('aep')).tap();
   await element(by.id('track2')).tap();
+
+  // Hackity hack to break up long URL.
```
jmtaber129: Let's add a TODO to remove this once the issue is fixed, and tag this comment with `:HACK:`, i.e. `// :HACK: Break up ...` (same with each of the delays below)

PR #66, opened 2019-04-06
android: use parent project's version information if available

natethebosch:
Fixes issue where parent project on a different sdkVersion wouldn't be able to build this library

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -36,6 +36,6 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    api "com.heapanalytics.android:heap-android-client:0.8.3"
+    api 'com.heapanalytics.android:heap-android-client:0.8.3'
 }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -19,7 +19,7 @@ android {
     compileSdkVersion safeExtGet('compileSdkVersion', 26)
 
     defaultConfig {
-        minSdkVersion safeExtGet('minSdkVersion', 16)
+        minSdkVersion 16
         targetSdkVersion safeExtGet('targetSdkVersion', 26)
         versionCode 1
         versionName "1.0"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,14 +9,18 @@ buildscript {
     }
 }
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion safeExtGet('compileSdkVersion', 26)
 
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 26
+        minSdkVersion safeExtGet('minSdkVersion', 16)
+        targetSdkVersion safeExtGet('targetSdkVersion', 26)
         versionCode 1
         versionName "1.0"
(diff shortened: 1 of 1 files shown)


GrandPoohBear:
Thanks @natethebosch !  We're going to get the right person on our team to take a look at this shortly :-)

android/build.gradle
```diff
@@ -31,6 +35,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.facebook.react:react-native:+'
-    api 'com.heapanalytics.android:heap-android-client:0.8.3'
+    implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
```
GrandPoohBear: Just curious (I am not a gradle guy at all...) - where does `reactNativeVersion` get set?  Is that something that will reliably be there on all installations?

natethebosch: To be honest, I've still got a lot to learn about gradle too... 
- reactNativeVersion comes from the dependencies section as far as I can tell.

I came across this when react-native-heap (and a couple other libraries) failed to compile for me. Looking at the build.gradle for some other libraries, all of them used `safeExtGet` for other dependencies. For example [react-native-device-info](https://github.com/react-native-community/react-native-device-info/blob/master/android/build.gradle).

reactNativeVersion won't always be set, but that's ok b/c `safeExtGet` will use the default value if it's not defined.

android/build.gradle
```diff
@@ -31,6 +35,7 @@ repositories {
...
+    api "com.heapanalytics.android:heap-android-client:0.8.3"
```
timothyandrew: nit: let's use single quotes here since this doesn't rely on interpolation.

android/build.gradle
```diff
@@ -9,14 +9,18 @@ buildscript {
...
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion safeExtGet('compileSdkVersion', 26)
 
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 26
+        minSdkVersion safeExtGet('minSdkVersion', 16)
```
timothyandrew: I don't think we want to inherit this value from the parent project; this would allow this library to be used by API versions below 16, which we don't currently support. I certainly see the value in inheriting the other two SDK version-related parameters, though!

android/build.gradle
natethebosch: 👍

android/build.gradle
natethebosch: That makes sense. What do you think about something like `safeExtGetWithMin('minSdkVersion', 16)` which does something like this:
```
if(!propExistsInParent) 
   return defaultAndMinValue
if(propInParent >= defaultAndMinValue) 
   return propInParent
return defaultAndMinValue
```

timothyandrew: I'm struggling to think of a reason `minSdkVersion` would need to be parameterized at all:

- If the parent project has `minSdkVersion` < 16, this library should (rightly) refuse to compile.
- If the parent project has `minSdkVersion` == 16, there are no issues.
- If the parent project has `minSdkVersion` > 16, that doesn't cause any issues either since this only specifies the minimum version.

I'm happy to let this conditional in if you have a concrete use case for it, but I'd prefer to avoid the complexity unless absolutely necessary.

natethebosch: I think you're right @timothyandrew, having that seems unnecessary when you think about it. I'll remove that.

timothyandrew: Sounds good, thank you! 👍

PR #69, opened 2019-04-10
Updates to the iOS `TestDriver` app.

timothyandrew:
These changes were created when I ran `pod install` in `TestDriver/iOS`. The main thing to verify here is whether these are legitimate changes or problems with my local setup.

### DIFF
--- a/examples/TestDriver/ios/Podfile.lock
+++ b/examples/TestDriver/ios/Podfile.lock
@@ -55,7 +55,7 @@ PODS:
     - React/Core
     - React/fishhook
     - React/RCTBlob
-  - RNGestureHandler (1.0.17):
+  - RNGestureHandler (1.1.0):
     - React
   - yoga (0.57.5.React)
 
@@ -105,11 +105,11 @@ SPEC CHECKSUMS:
   DoubleConversion: bb338842f62ab1d708ceb63ec3d999f0f3d98ecd
   Folly: c89ac2d5c6ab169cd7397ef27485c44f35f742c7
   glog: e8acf0ebbf99759d3ff18c86c292a5898282dcde
-  React: 1fe0eb13d90b625d94c3b117c274dcfd2e760e11
-  react-native-heap: 381fd78115a2a1314dae2906b893221c9501d12d
-  RNGestureHandler: afde2addc517e85cc97a25bfe119acbee779a7ce
-  yoga: b1ce48b6cf950b98deae82838f5173ea7cf89e85
+  React: 01aa04500b2957c2767e1dff9fe12e09444a467c
+  react-native-heap: e75c689169e157e25a236be2c072bf24aa195824
+  RNGestureHandler: b65d391f4f570178d657b99a16ec99d09b8656b0
+  yoga: a5c0ba30ebe82c13612b4c9301ad29708b8978dc
 
 PODFILE CHECKSUM: 94963ca9da08a6f38a9019eb8da64ec0f890b712
 
-COCOAPODS: 1.5.3
+COCOAPODS: 1.6.1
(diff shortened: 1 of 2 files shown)


examples/TestDriver/ios/Podfile.lock
```diff
@@ -55,7 +55,7 @@ PODS:
     - React/Core
     - React/fishhook
     - React/RCTBlob
-  - RNGestureHandler (1.0.17):
+  - RNGestureHandler (1.1.0):
```
GrandPoohBear: Ah, the version of RNGestureHandler (as dictated by `npm`, since we're just referring to the `.podspec` in `node_modules`) changed.  I remember we had some weirdness around that not being there, and we had added it explicitly at some point.  Not a big deal - feel free to check this in!

examples/TestDriver/ios/Podfile.lock
```diff
@@ -105,11 +105,11 @@ SPEC CHECKSUMS:
   DoubleConversion: bb338842f62ab1d708ceb63ec3d999f0f3d98ecd
   Folly: c89ac2d5c6ab169cd7397ef27485c44f35f742c7
   glog: e8acf0ebbf99759d3ff18c86c292a5898282dcde
-  React: 1fe0eb13d90b625d94c3b117c274dcfd2e760e11
-  react-native-heap: 381fd78115a2a1314dae2906b893221c9501d12d
-  RNGestureHandler: afde2addc517e85cc97a25bfe119acbee779a7ce
-  yoga: b1ce48b6cf950b98deae82838f5173ea7cf89e85
+  React: 01aa04500b2957c2767e1dff9fe12e09444a467c
```
GrandPoohBear: Hashes changed - not too worried since we're really worried about tracking versions in npm rather than Cocoapods.  I wonder if there's a way to configure Cocoapods to do its thing without a lock file?  In this case, it just gets in our way, I think.

examples/TestDriver/ios/Podfile.lock
```diff
@@ -105,11 +105,11 @@ SPEC CHECKSUMS:
...
+  react-native-heap: e75c689169e157e25a236be2c072bf24aa195824
+  RNGestureHandler: b65d391f4f570178d657b99a16ec99d09b8656b0
+  yoga: a5c0ba30ebe82c13612b4c9301ad29708b8978dc
 
 PODFILE CHECKSUM: 94963ca9da08a6f38a9019eb8da64ec0f890b712
 
-COCOAPODS: 1.5.3
+COCOAPODS: 1.6.1
```
GrandPoohBear: So this is one of those pain in the ass things, because you have a different version of Cocoapods than I do, clearly, and I didn't even realize there was an update available... We ran into this problem at Udacity, and it's just a small burden that dev teams take on, unless you start standardizing your ruby gems etc. which is a whole different PITA...

examples/TestDriver/ios/TestDriver.xcodeproj/project.pbxproj
```diff
@@ -393,7 +393,7 @@
 			inputFileListPaths = (
 			);
 			inputPaths = (
-				"${SRCROOT}/Pods/Target Support Files/Pods-TestDriver/Pods-TestDriver-frameworks.sh",
+				"${PODS_ROOT}/Target Support Files/Pods-TestDriver/Pods-TestDriver-frameworks.sh",
```
GrandPoohBear: I'm guessing this is a byproduct of the different CP version - looks like they're probably equivalent when expanded.

timothyandrew:
Sounds good! I'm merging this in now; feel free to revert if this ends up causing any trouble.

PR #70, opened 2019-04-16
[HEAP-8473] Capture the `key` property for list items that have it set.

timothyandrew:
- We have to special-case this a bit because React doesn't allow access to
the `key` property without a yellowbox warning popping up.

- e2e tests are passing locally.

https://reactjs.org/docs/lists-and-keys.html

### DIFF
--- a/js/util/extractProps.ts
+++ b/js/util/extractProps.ts
@@ -15,6 +15,7 @@ export interface FiberNode {
     [propertyKey: string]: any;
   };
   stateNode?: StateNode;
+  key?: any;
   type?: {
     heapOptions?: ClassHeapOptions;
   };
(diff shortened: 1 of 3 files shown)


js/util/extractProps.ts
```diff
@@ -87,7 +88,14 @@ export const extractProps = (
...
 
   const filteredProps = _.pick(props, inclusionList);
-  const flattenedProps = flatten(filteredProps);
+
+  // KLUDGE: We want to capture the `key` property for list components that have it set,
+  // but we can't simply add to the list of props captured for all components in
+  // `propExtractorConfig` because `props.key` is always undefined and guarded with
+  // a yellowbox warning; this is to prevent client code from appropriating the `key`
+  // prop for its own use; it is intended to be reserved for internal use. (HEAP-8473)
+  const flattenedProps = Object.assign({ key: fiberNode.key }, flatten(filteredProps));
```
GrandPoohBear: Nice catch!

FWIW, I've always thought the splat syntax made this nice to read, but your call!
```
const flattenedProps = { ...flatten(filteredProps), key: fiberNode.key }
```

timothyandrew: Ah, didn't realize this was possible with objects. Looks like it isn't included in ES6 and needs a babel plugin, though, so I think I'll just leave it the way it is for now.

https://redux.js.org/recipes/using-object-spread-operator

timothyandrew:
I'd suggest improving the testing harness (to log/diff the actual values that differ in a failing test) rather than adding a separate test case here; unless there are specific edge cases (I can't think of any), I don't think they'd add any value.

PR #72, opened 2019-04-19
Refactor RN test utils to accept generic props

jmtaber129:
Currently, our RN test utils are hard coded to assert the props `touchableHierarchy` and `targetText`.  This doesn't work when these props are different or not present (e.g. `targetText` is not a prop for `Switch` components, since these don't have any inner text).

Instead, we should take a generic `expectedProps` object, and expand this to the correct format for ios/android when making the event assertion.

~Note: this has been tested on iOS, but not on android (having trouble with my local android build for TestDriver), and so should not yet be merged~.  This has now been successfully tested on android

### DIFF
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -228,43 +228,41 @@ describe('Basic React Native and Touchable Support', () => {
         'AppContainer;|App;|Provider;|HeapNavigationWrapper;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;[key=Basics];|SceneView;|Connect(BasicsPage);|BasicsPage;|TouchableOpacity;[testID=touchableOpacityText];|';
       const expectedTargetText = 'Touchable Opacity Foo';
       await rnTestUtil.assertAutotrackHierarchy(
-        'touchableHandlePress',
-        expectedHierarchy,
-        expectedTargetText
-      );
+      'touchableHandlePress',
+        {
+          touchableHierarchy: expectedHierarchy,
+          targetText: expectedTargetText,
+        });
     });
 
     it("should autotrack 'TouchableHighlight's", async () => {
       const expectedHierarchy =
         'AppContainer;|App;|Provider;|HeapNavigationWrapper;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;[key=Basics];|SceneView;|Connect(BasicsPage);|BasicsPage;|TouchableHighlight;[testID=touchableHighlightText];|';
       const expectedTargetText = 'Touchable Highlight';
-      await rnTestUtil.assertAutotrackHierarchy(
-        'touchableHandlePress',
-        expectedHierarchy,
-        expectedTargetText
-      );
+      await rnTestUtil.assertAutotrackHierarchy('touchableHandlePress', {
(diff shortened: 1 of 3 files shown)


e2e/rnTestUtilities.js
```diff
@@ -33,49 +36,30 @@ const assertAndroidEvent = async (event, check) => {
...
   } else if (device.getPlatform() === 'ios') {
     return assertIosPixel({
       t: expectedName,
-      k: [
-        'touchableHierarchy',
-        expectedHierarchy,
-        'targetText',
-        expectedTargetText,
-      ],
+      k: _(expectedProps)
```
GrandPoohBear: I think I see what you're doing here, but this is just clever enough to warrant an explanation :-)

e2e/rnTestUtilities.js
```diff
@@ -33,49 +36,32 @@ const assertAndroidEvent = async (event, check) => {
...
-        'touchableHierarchy',
-        expectedHierarchy,
-        'targetText',
-        expectedTargetText,
-      ],
+      // Convert { key1: 'value1', key2: 'value2'} to ['key1', 'value1', 'key2', 'value2'] for
+      // custom props.
+      k: _(expectedProps)
+        .map((value, key) => [key, value])
+        .flatten()
```
timothyandrew: Could this be `_.flatMap(expectedProps, (v,k) => [k,v])` instead?

e2e/basics.spec.js
```diff
@@ -228,43 +228,41 @@ describe('Basic React Native and Touchable Support', () => {
         'AppContainer;|App;|Provider;|HeapNavigationWrapper;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;|SceneView;|Connect(BasicsPage);|BasicsPage;|TouchableOpacity;[testID=touchableOpacityText];|';
       const expectedTargetText = 'Touchable Opacity Foo';
       await rnTestUtil.assertAutotrackHierarchy(
-        'touchableHandlePress',
-        expectedHierarchy,
-        expectedTargetText
-      );
+      'touchableHandlePress',
```
timothyandrew: nit: spacing seems off here.

GrandPoohBear: With `prettier`, we should never have to worry about that thing.  If we want something different, we need to build it into the formatter rules!

e2e/rnTestUtilities.js
jmtaber129: TIL about `_.flatMap`! 👌

e2e/basics.spec.js
jmtaber129: Yea, I'm also not a huge fan of our current prettier formatting rules, but it's better than inconsistent formatting and having to manually format everything

jmtaber129:
@timothyandrew's review is technically "requested changes", but includes "LGTM" in the review body, and the minor comments have been resolved.  AFAIK he's signed off for the day, so I'm force-merging this PR via admin privileges.

timothyandrew:
> @timothyandrew's review is technically "requested changes", but includes "LGTM" in the review body, and the minor comments have been resolved. AFAIK he's signed off for the day, so I'm force-merging this PR via admin privileges.

Thanks, this was definitely my intention; Github doesn't have a nice way to encode "approved once these changes are addressed" in the review workflow.

PR #73, opened 2019-04-22
DO NOT MERGE: [HEAP-8587] Fix sessionization for the `TestDriver` android app.

timothyandrew:
It turns out that `Heap.init` is called (from Javascript) too late,
after `MainActivity` is created. This leads our library to assume that
`MainActivity` is backgrounded, and so all events are treated as custom
events, with each receiving a different session ID.

The fix is simple; call `Heap.init` from `MainActivity` directly (in
Java), before the Javascript is loaded.

---

- I attempted to find a way to have the Java code in our library automate this process, but I didn't get too far with this. Ideas welcome! The only approach I can think of is to use our redex instrumentor to add `Heap.setAppId` to `MainActivity`, which definitely seems like overkill here.
- Once this approach / API is approved, I'll:
    - [ ] Update the READMEs in the repo and docs.heap.io so they contain this extra step.
    - [ ] Add a warning if js-level `Heap.setAppId` is called without java-level `Heap.setAppId` being called first.


kamalmarhubi:
> LGTM, but I'm curious about the consequences of calling Heap.init both in the native code and from the react code?

We log "already initialised" but we made this a noop otherwise. However, we do not expose a way to change the env id. I haven't looked in detail at installation instructions, but based on [the Usage section in README](https://github.com/heap/react-native-heap#usage), that seems like a show stopper?

timothyandrew:
https://heap.slack.com/archives/CFGQWRN56/p1556043025022600

PR #77, opened 2019-04-23
E2E tests for RN Switch component

jmtaber129:
This PR includes:
* e2e tests for switches
* a fix for a bug that was uncovered by the tests

Specifically, the bug occurred when the bundler generated bundles for both platforms within the same process.  Because the `instrumentedComponentNodes` set is not re-initialized for each new platform, the following happens:
* generate bundle for iOS
* mark 'Switch' as visited in `instrumentedComponentNodes`
* generate bundle for android
* 'Switch' is contained in `instrumentedComponentNodes`, so don't instrument `Switch`

By adding the `path` object to the set, rather than the string representation 'Switch', the key will be different between platforms, but still the same for subsequent visits to the same node within the same platform (i.e. keys will be different between iOS and android, but the key will be the same when the node is visited multiple times within a single iOS traversal).

### DIFF
--- a/examples/TestDriver/src/pages/basics.js
+++ b/examples/TestDriver/src/pages/basics.js
@@ -22,7 +22,7 @@ class BasicsPage extends Component {
   render() {
     return (
       <View style={styles.container}>
-        <NbSwitch testID="switch" onValueChange={() => console.log('stuff')} />
+        <NbSwitch testID="nbSwitch" onValueChange={() => console.log('stuff')} />
         <Switch testID="switch" onValueChange={() => console.log('stuff')} />
         <Button
           testID="track1"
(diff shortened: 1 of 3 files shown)


examples/TestDriver/src/pages/basics.js
```diff
@@ -22,7 +22,10 @@ class BasicsPage extends Component {
   render() {
     return (
       <View style={styles.container}>
-        <NbSwitch testID="switch" onValueChange={() => console.log('stuff')} />
+        <NbSwitch
+          testID="nbSwitch"
+          onValueChange={() => console.log('stuff')}
```
GrandPoohBear: nit - cruft?

PR #78, opened 2019-04-23
[HEAP-8830] Don't allow errors thrown by the Heap library to crash the client app.

timothyandrew:
- Errors are swallowed in production and surfaced in debug builds.
- This covers the public API and the autotracking code.

### DIFF
--- a/js/Heap.js
+++ b/js/Heap.js
@@ -6,11 +6,13 @@ import { autotrackPress } from './autotrack/touchables';
 import { autotrackSwitchChange } from './autotrack/switches';
 import { checkDisplayNamePlugin } from './util/checkDisplayNames';
 import { withReactNavigationAutotrack } from './autotrack/reactNavigation';
+import { bailOnError } from './util/bailer';
 
 const flatten = require('flat');
 const RNHeap = NativeModules.RNHeap;
 
-const track = (event, payload) => {
+
+const track = bailOnError((event, payload) => {
   try {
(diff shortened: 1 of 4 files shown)


timothyandrew:
Good point, I'll look into this. There are a few other things pending as well before this is ready for a full review:

- Allow crashes in debug builds, either by re-throwing the error or displaying a warning.
- Potentially disable tracking (with backoff) on crash.

timothyandrew:
This should be ready for review. I added a second commit with a test covering some of the crash handling behavior, but this requires adding a method to the public `Heap` API, and I'm not sure if that's worth it. Thoughts?

js/Heap.js
```diff
@@ -23,33 +25,40 @@ const track = (event, payload) => {
   } catch (e) {
     console.log('Error calling Heap.track\n', e);
   }
+});
+
+const triggerCrash = () => {
```
GrandPoohBear: Is there a reason we need this in the actual library code vs. the test code?  I'm assuming this is a _test only_ function...

Also, I think it'd probably be funny if I was looking in the source of a new lib and the first thing I saw was a `triggerCrash` function!!!

timothyandrew: Good point! I wasn't able to find a way to add the `triggerCrash` function to the `Heap` API from the detox level, so I simply moved to unit tests with Jest instead. These probably cover more of the actual code involved anyway, so I'd say this was for the best.

PR #82, opened 2019-04-25
[HEAP-8938] Auto-initialize Heap for React Native+Android apps.

timothyandrew:
- Expect (but not require) a `heap.json` file at the root of client
projects that contains the appId. Auto-initialization can also be
disabled using a flag in this file.

- Auto-initialization is nice to have, but is required to fix
sessionization issues on RN+Android builds; the previous approach
(`Heap.setAppId` from Javascript) executed too late, causing the native
Android library to assume that the main activity was in the background
when it was not.

- No action is performed when auto-initialization fails, which can
happen because:

    * `heap.json` doesn't exist.
    * `heap.json` doesn't contain an appId.
    * `heap.json` explicitly disables auto-initialization.  

- I plan to update the documentation at
https://docs.heap.io/docs/react-native to include manual initialization
instructions (in `MainActivity`) if a customer really wants to do it
themselves.

- JSON is typically not a great choice for configuration files; it
doesn't support comments, it is a bit verbose, etc. However in this
case, I'd argue that it happens to be the best fit because it is the
only serialization format that both Groovy and Ruby (on the iOS side)
support without having to install any libraries. Happy to write a mini-RFC
about this if anyone has strong feelings against.


android/build.gradle
```diff
@@ -1,3 +1,31 @@
...
+def warnNoAutoInit() {
+    logger.warn("Couldn't auto-initialize the Heap library; manually initialize it by following the instructions at https://docs.heap.io/docs/react-native.")
+}
+
+def warnDisabled() {
+    logger.warn("Not auto-initializing the Heap library because it was manually disabled; set `enabled` to `true` in `heap.json` to enable auto-initialization.")
+}
+
+def heapAppId() {
+    def heapConfigFilename = 'heap.json'
```
GrandPoohBear: Could this be a CONSTANT at the top of the file?

android/build.gradle
```diff
@@ -1,3 +1,31 @@
...
+    def file = rootProject.file("../${heapConfigFilename}")
+    if (!file.exists()) { return warnNoAutoInit() }
+
+    def jsonSlurper = new JsonSlurper()
+    def data = jsonSlurper.parseText(file.text.trim())
+
+    if (!data instanceof Map || !data.appId instanceof String) { return warnNoAutoInit() }
```
GrandPoohBear: I think this gets a bit funny to read... might be nice to phrase this in the positive rather than the negative (especially considering there will just be a single `return` inside the conditional block.

Which, BTW, I've always followed style guidelines to break those blocks into three lines (i.e. "{\nFN_BODY\n}") for readability, except in a few rare circumstances.  It seems weightier on the page, and draws my eye into it.

That being said, I'm a big fun of the `guard let ...` construct common in Swift, where an early abort is just done on a single line... so you commonly see things like:
```
guard let neverNullVal = possiblyNullVal else { return }
```
... which is also nice because you've got a variable forever after that line that's guaranteed to be good and you don't have to worry about it.  And in that case, I think the braces on the same line as the conditional makes things neat and tidy, so go figure.

What do you think?

android/build.gradle
```diff
@@ -23,6 +47,15 @@ android {
         targetSdkVersion safeExtGet('targetSdkVersion', 26)
         versionCode 1
         versionName "1.0"
+
+        def appId = heapAppId()
+        if (appId) {
+            buildConfigField("String", "HEAP_APP_ID", "\"${appId}\"")
```
GrandPoohBear: `"String"` and `"HEAP_APP_ID"` should probably be pulled out into a constant (more so the latter than the former, I think)... the goal would be just to reduce the number of accidental mistakes someone could make by fat-fingering a known string as they type it in... both for yourself (and I think we can both clearly see that you got it right) and for the next person!

android/src/main/java/com/heapanalytics/reactnative/HeapInitializationProvider.java
```diff
@@ -0,0 +1,42 @@
+package com.heapanalytics.reactnative;
```
GrandPoohBear: Hey @kamalmarhubi - any thoughts about changing the package name to match the company name (i.e. drop the analytics)?  Are we locked into that for weird reasons?

e2e/basics.spec.js
```diff
@@ -151,6 +152,17 @@ describe('Basic React Native and Interaction Support', () => {
       );
     });
 
+    it('preserve the same session between track calls', async () => {
```
GrandPoohBear: Nice!

examples/TestDriver/heap.json
```diff
@@ -0,0 +1,4 @@
+{
+    "appId": "2084764307",
+    "enabled": true
```
GrandPoohBear: I think this key needs to be a little more descriptive... perhaps `autoInitEnabled` or something?  If I stumbled on this file, I could have a lot of interpretations of what `enabled` means!

Also, I could see us using this file for a lot of things... maybe we want to start allowing configurable prop capture in here like we do for the globals?  Then it wouldn't need to be added to _code_ per se.

@jnatkins Want to make sure I flag this for you - this will be a non-breaking change, but something we should keep current users up to date on and revise docs etc.

kamalmarhubi: over in android land, the (as-yet unofficial) names of these variables are: `heapEnvId`, `heapEnabled`, `heapAutoInit`. The `heap` prefixes can go in this world of `heap.json`, and `enabled` is actually "run instrumentation" so isn't relevant here. So `autoInit`?

android/build.gradle
timothyandrew: I attempted doing this for about 5 minutes but it didn't work for some reason. I'm not sure if Gradle+Groovy has constants per se, but I could just move this `def` up to the global scope I suppose.

jnatkins:
One high-level comment I would make is that customers sometime parameterize the Heap app ID. They do this to make it easier to switch between dev and prod Heap apps, that are driven by the release flag on the RN build. If there’s a way to do that with auto-init, it’s ideal, but we should optimize for fixing the sessionization.

examples/TestDriver/heap.json
jnatkins: I’d strongly advise having the variables named consistently, if it’s not a ton of work to refactor. It makes work easier for docs, and allows customers to retain their mental model.

android/build.gradle
timothyandrew: My intuition was that `resValue`s are more for UI-visible text/etc., while `BuildConfig` fields are for internal dispatch. Why/when would you pick one of these over the other?

android/build.gradle
```diff
@@ -1,3 +1,38 @@
+import groovy.json.JsonSlurper
+
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
+def warnNoAutoInit() {
+    logger.warn("Couldn't auto-initialize the Heap library; manually initialize it by following the instructions at https://docs.heap.io/docs/react-native.")
```
GrandPoohBear: Could we add a "please"?  LOL... I'm such a dad of toddlers...

android/build.gradle
```diff
@@ -1,3 +1,38 @@
...
+    def file = rootProject.file("../${HEAP_CONFIG_FILENAME}")
+    if (!file.exists()) { return warnNoAutoInit() }
+
+    def jsonSlurper = new JsonSlurper()
+    def data = jsonSlurper.parseText(file.text.trim())[buildType]
+    def defaultData = jsonSlurper.parseText(file.text.trim())["default"]
+
+    // KLUDGE: Can't use the elvis operator because we want to treat `null` and `false` differently.
+    def autoInit = data?.heapAutoInit
+    if (autoInit == null) { autoInit = defaultData?.heapAutoInit }
```
GrandPoohBear: It's really not clear to me what's going on here and the rest of the function ... now granted, it's been a while since I've gotten groovy, but I suspect there's something you could do to reorganize the conditions a bit, for clarity.  Also, is the repeated `autoInit == null` a mistake?

android/src/main/java/com/heapanalytics/reactnative/HeapInitializationProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+        if (!appId.isEmpty()) {
+            RNHeap.init(getContext(), appId);
+        }
+
+        return true;
+    }
+
+    @Override
+    public Cursor query(Uri uri, String[] strings, String s, String[] strings1, String s1) {
```
GrandPoohBear: Wow, this is an unfortunate API... I realize it's not yours, so not like you can change it, but ick... is this just like pre-collections java or did string arrays come back into favor since I left?

android/src/main/java/com/heapanalytics/reactnative/RNHeap.java
```diff
@@ -0,0 +1,12 @@
+package com.heapanalytics.reactnative;
+
+import android.content.Context;
+import com.heapanalytics.android.Heap;
+import com.heapanalytics.android.internal.HeapImpl;
+
+public class RNHeap {
+    public static void init(Context context, String appId) {
```
GrandPoohBear: This is interesting!  It's a new file, right?  Do all classes that expose an `init` method with this sig get called on app initialization or something?  Is this magic, or are we calling it explicitly?  Just super curious!  (I know, I know, I "wrote" the stuff to begin with, but believe me FAR from an Android maven)

android/src/main/java/com/heapanalytics/reactnative/RNHeap.java
```diff
@@ -0,0 +1,12 @@
...
+        HeapImpl.skipInstrumentorChecks = true;
```
GrandPoohBear: Also, is there somewhere else that we call this already from the RN Plugin?  It's totally idempotent to do it, but probably good to find any other spot where we do it already for hygiene sake.

android/build.gradle
timothyandrew: Suggestions? This seemed fairly readable to me; the intention essentially is: if `heapAutoInit` for the current build type is set, use it; if not (it's null), and `heapAutoInit` for the `default` build type is set, use it; if not (it's null too), just use `true`.

android/src/main/java/com/heapanalytics/reactnative/RNHeap.java
timothyandrew: It's a new file, yes, because I didn't want to put code that wasn't called from JS-land in the existing `RNHeapLibraryModule` class. Also this is not magic, we call this function from the context provider (HeapInitializationProvider.java).

android/src/main/java/com/heapanalytics/reactnative/RNHeap.java
timothyandrew: Yep, in `RNHeapLibraryModule.java`: https://github.com/heap/react-native-heap/blob/037e2e3cfed8d23ba7334dd2982870ab7c9619a9/android/src/main/java/com/heapanalytics/reactnative/RNHeapLibraryModule.java#L34

Not sure I understand the suggestion here.

GrandPoohBear: Just saying that we should only have the call in one spot.  It doesn't matter if we call it twice (idempotent), but, conceptually, this kind of initialization stuff should only happen once in our code.

android/build.gradle
GrandPoohBear: So I think I usually think of it upside down from what you've got.  Here's what I came up with, and I think they're equivalent (let me know if not!)
```groovy
    def autoInit = true // Default value
    if (data?.heapAutoInit != null) {
      autoInit = data?.heapAutoInit
    } else if (defaultData?.heapAutoInit != null) {
      autoInit = defaultData.heapAutoInit
    }
```
... definitely a bit more verbose, but I think a bit clearer.  It just _looks_ like an easily-made error from 5000 ft to have that same condition repeated twice - you have to actually grok the rest of the line to realize that the variable in the condition got modified in between those checks!

android/build.gradle
```diff
@@ -1,3 +1,38 @@
...
+    if (autoInit == null) { autoInit = true }
+
+    if (!autoInit) { return warnDisabled() }
+    return data?.heapAppId ?: defaultData?.heapAppId ?: warnNoAutoInit();
```
GrandPoohBear: Now that I'm looking at it again, this seems a little smelly... like we're either returning an app id (string) or we're... printing a warning??  Halting in the middle of the process??  Is there an exception that should be thrown or some other way to indicate what's gone wrong?

android/build.gradle
GrandPoohBear: Could even do something like (completely spitballing, haven't tried it):
```groovy
    def autoInit = true // Default value
    def configValue = null
    if ((configValue = data?.heapAutoInit) != null) {
      autoInit = configValue
    } else if ((configValue = defaultData?.heapAutoInit) != null) {
      autoInit = configValue
    }
```

android/src/main/java/com/heapanalytics/reactnative/HeapInitializationProvider.java
```diff
@@ -0,0 +1,50 @@
...
+        String appId;
+
+        if (BuildConfig.DEBUG) {
+            appId = BuildConfig.HEAP_APP_ID_DEV;
+        } else {
+            appId = BuildConfig.HEAP_APP_ID_PROD;
+        }
+
+        if (!appId.isEmpty()) {
+            RNHeap.init(getContext(), appId);
```
GrandPoohBear: Ah, I see... so this init pathway will ONLY get hit if we HAVE a valid BuildConfig.  Never mind about my comment about unifying those two calls then!

android/src/main/java/com/heapanalytics/reactnative/RNHeap.java
GrandPoohBear: Nvm my last comment... I get it now :-P

android/build.gradle
timothyandrew: Ack. These versions seem _less_ readable to me, but I have been looking at this for a while, so your impression is likely a lot more accurate than mine. 🙂

android/src/main/java/com/heapanalytics/reactnative/HeapInitializationProvider.java
timothyandrew: Not too sure off the top of my head; the only reason I can think of is performance: maybe they're trying to avoid the extra overhead of using function calls to interact with the collection? I don't really see any other reason to prefer `String[]` over `List<String>`.

android/build.gradle
timothyandrew: The contract here is that a `null` return value indicates that an app id could not be found. I'm not a big fan of using exceptions for predictable errors.

timothyandrew:
@GrandPoohBear: I'm merging this now; let me know if there's anything here you want to re-address before the next release. I've confirmed that E2E tests pass locally.

PR #83, opened 2019-04-30
Auto-initialize Heap for React Native on iOS (Cocoapods only)

timothyandrew:
RFC: https://paper.dropbox.com/doc/RFC-React-Native-Configuration-File--Ab9Qxy0sEh62Ud3ZRCeUFvafAg-VplkwaTstFyBAEBF5ZEdT
Android implementation: https://github.com/heap/react-native-heap/pull/82


react-native-heap.podspec
```diff
@@ -3,6 +3,52 @@ require "json"
...
+    app_id_dev = dev[:heapAppId] || default[:heapAppId] if auto_init_dev
+    app_id_prod = prod[:heapAppId] || default[:heapAppId] if auto_init_prod
+  else
+    puts "Not auto-initializing Heap; couldn't find #{HEAP_CONFIG_FILENAME}."
+  end
+
+  template = File.read(File.join(__dir__, 'ios', 'HeapSettings', 'HeapSettings.plist.template'))
+  output_path = File.join(__dir__, 'ios', 'HeapSettings', 'HeapSettings.plist')
+
+  settings = template
```
GrandPoohBear: I would suggest using `plistbuddy` for something like this.  See, e.g., https://willowtreeapps.com/ideas/using-plistbuddy-to-inject-app-secrets-and-keys-at-build-time for a very similar use case.

timothyandrew:
This is good feedback! I did give the "Build Phase" approach a try, but I wasn't able to get a "Build Phase" defined in our library running for the client app with Cocoapods enabled (didn't try this with the non-Cocoapods approach); do you have a sense about whether or not this is possible?

I'll give this another shot today; if I don't get anywhere, let's pair and figure this out. 👍

timothyandrew:
@GrandPoohBear: I added a commit that uses CocoaPods to add a build phase that generates the plist containing the app id on every build. This is _much_ nicer, because (like you said) a `pod install` is not required after every configuration change. A few notes:

- The script itself is a bit hacky, but not more so than the previous approach. Curious to hear your thoughts.

- Adding the build script directly to XCode didn't work, presumably because CocoaPods ignores/overwrites it. I haven't tried this yet, but I presume that we can use this approach for non-CocoaPods builds.

- Comments in the code should make this easy to understand, but let me know if you have any questions.

PR #84, opened 2019-05-02
Auto-initialize Heap for RN+iOS apps that aren't using CocoaPods.

timothyandrew:
- This is marginally different from the CocoaPods version in #83, mainly the specific paths to the `HeapSettings` bundle.

- This adds an extra manual step for non-CocoaPods setup; the setup for non-CocoaPods apps already requires manual steps, so I didn't see the need to spend time automating this in some way.

- I tested on this by adding Heap autotrack to this app: https://github.com/7kfpun/FinanceReactNative.


ios/HeapSettings.bundle/generate_settings_non_cocoapods
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env ruby
```
GrandPoohBear: I this this file should also have a `.rb` extension

timothyandrew:
Thanks for the super quick review! I'll run tests locally and merge this thing.

PR #86, opened 2019-05-09
[HEAP-9116] Implement basic HeapIgnore functionality

jmtaber129:
Implement HeapIgnore ~and HeapCaptureRestrictor (may not make it into merged thing)~ component.  Does *not* include the `ignoreSpecificProps` config prop due to added complexity that will be included in a future PR.

The components themselves are just placeholders that just accept props and wrap the component children.  When we traverse the hierarchy, we look for these HeapIgnore components, and use the props to modify the string representation of the subhierarchy, as well as store the state to later determine whether to ignore target text or the entire event.

Part of this PR is refactoring how we traverse the component hierarchy.  Currently, we have a single step - recursively traverse up the hierarchy from target element to the root, while assembling the hierarchy string.  To support HeapIgnore, this is being changed to two steps:
* Recursively traverse up the hierarchy to the root, while assembling an array containing relevant per-component data representing the overall hierarchy.
* Iterate over the array, assembling the hierarchy string from root to target element, while recomputing the current state of HeapIgnore props for the current hierarchy subtree.

~For now, just want comments on general approach (will want style comments/etc. after I clean up a bit more and add tests). Leaving this as a draft until I add tests and clean up a little more.~


examples/TestDriver/src/pages/basics.js
```diff
@@ -14,63 +14,74 @@ import {
 import { Switch as NbSwitch } from 'native-base';
 import { connect } from 'react-redux';
 
-import Heap from '@heap/react-native-heap';
+import Heap, { HeapIgnore, HeapCaptureRestrictor } from '@heap/react-native-heap';
 import { incrementAction, decrementAction } from '../reduxElements';
 import { makeSentinelButton } from '../sentinelUtilities';
 
+const HeapIgnoredNbSwitch = Heap.withHeapIgnore(NbSwitch, {
+  ignoreInteraction: false,
```
GrandPoohBear: I think I know what you're saying here, but it's a funky name.  Perhaps `suppressAllTouchEvents` or something?  Is it limited to touch events?  I think so, right?

examples/TestDriver/src/pages/basics.js
```diff
@@ -14,63 +14,74 @@ import {
...
+  ignoreAllProps: true,
+  ignoreTargetText: false,
+  ignoreInnerHierarchy: false,
```
GrandPoohBear: At first glance, no idea what this means...

examples/TestDriver/src/pages/basics.js
```diff
@@ -14,63 +14,74 @@ import {
...
-          testID="removeProp"
-          title="Remove eventProp1"
-          onPress={() => Heap.removeEventProperty('eventProp1')}
-        />
-        <Button
-          testID="clearProps"
-          title="Clear Event Properties"
-          onPress={() => Heap.clearEventProperties()}
-        />
+        <HeapCaptureRestrictor ignoreTargetText={true}>
```
GrandPoohBear: High level - what's the semantic difference between `HeapCaptureRestrictor` and `HeapIgnore`?  My first thought looking at this is that I'm not sure why they don't do the same thing...

js/Heap.js
```diff
@@ -21,11 +22,15 @@ const track = bailOnError((event, payload) => {
...
 
     checkDisplayNamePlugin();
+
+    console.log('Heap.track call:', event, payload);
   } catch (e) {
     console.log('Error calling Heap.track\n', e);
   }
 });
 
+export { HeapIgnore, HeapCaptureRestrictor };
```
GrandPoohBear: I like the way you set this up... Could I actually use this as `<Heap.Ignore ...` or does that not work?  I think I like the way that looks slightly better, but I also like that it could be accessed either way.

js/autotrack/common.js
```diff
@@ -1,13 +1,31 @@
...
 
+// Returns an object containing a base set of component properties if we're not ignoring the
+// full interaction due to HeapIgnore.
+// Returns null if we're ignoring the full interaction due to HeapIgnore.
 export const getBaseComponentProps = componentThis => {
-  const touchableHierarchy = getComponentHierarchy(componentThis);
+  // Get the hierarchy traversal from root to target component, then get the actual hierarchy from
+  // the traversal representation.
+  const touchableHierarchyTraversal = getComponentHierarchyTraversal(componentThis);
+  const { hierarchy, heapIgnoreProps } = getHierarchyStringFromTraversal(touchableHierarchyTraversal);
```
GrandPoohBear: Nice, clean...

js/autotrack/common.js
```diff
@@ -1,13 +1,31 @@
...
 import { builtinPropExtractorConfig } from '../propExtractorConfig';
 
+// Returns an object containing a base set of component properties if we're not ignoring the
+// full interaction due to HeapIgnore.
+// Returns null if we're ignoring the full interaction due to HeapIgnore.
 export const getBaseComponentProps = componentThis => {
-  const touchableHierarchy = getComponentHierarchy(componentThis);
+  // Get the hierarchy traversal from root to target component, then get the actual hierarchy from
+  // the traversal representation.
+  const touchableHierarchyTraversal = getComponentHierarchyTraversal(componentThis);
```
GrandPoohBear: I'm a bit confused what a "Traversal" is in this sense.  I get that you have to go all the way up and then come back down to cover the ignored bits, but has it materially changed in terms of what you're getting out of the function call?

GrandPoohBear: And lol, yes, this question would be MUCH easier to answer in typescript... I'd be in favor of just ripping the band-aid and making this ts happy.

I also think, btw, that the code in this file predates when we started trying ts, so it's possible there's no good reason at all!

js/autotrack/common.js
```diff
@@ -57,11 +81,104 @@ const getFiberNodeComponentHierarchy = currNode => {
...
+    if (currentHeapIgnoreProps.ignoreInteraction || currentHeapIgnoreProps.ignoreInnerHierarchy) {
+      return '';
+    }
+
+    if (currentHeapIgnoreProps.ignoreAllProps) {
+      return `${element.elementName};|`;
+    }
+
+    // Doing this at the end allows us to capture HeapIgnore components.
+    currentHeapIgnoreProps = getNewHeapIgnoreProps(currentHeapIgnoreProps, element);
```
GrandPoohBear: Why are they new?  Is there something more descriptive we could use there?

js/autotrack/touchables.js
```diff
@@ -1,16 +1,20 @@
 import { getBaseComponentProps } from './common';
 
 export const autotrackPress = track => (eventType, componentThis, event) => {
+  const autotrackProps = getBaseComponentProps(componentThis);
```
GrandPoohBear: Seems like it's a common occurrence.  Maybe worth pulling that out in a way we can have all these functions just pick it up?  Nothing comes to mind, but maybe you have ideas...

examples/TestDriver/src/pages/basics.js
jmtaber129: `HeapIgnore` ignores _everything_ by default (i.e. you'd have to explicitly pass `ignoreInteraction={false}` if you want to capture _anything_).  I.e. it's best if you want to ignore some piece of the hierarchy.

`HeapCaptureRestrictor` captures everything by default (i.e. if you passed `ignoreTarget={true}` it would restrict capture to everything that isn't target text).  I.e. it's best if you want to restrict what we capture to some extent.

My thinking is that there would be cases where it's only "safe" to capture only the fact that there was any interaction with a component (if it's safe to capture anything), but there might also be cases where you just want to leave out target text and props for a few buttons.

js/Heap.js
jmtaber129: Yea, the idea is that 
```
import { HeapIgnore } rom '@heap/react-native-heap';
...
<HeapIgnore>
```
```
import Heap rom '@heap/react-native-heap';
...
<Heap.Ignore>
```
are the same thing.

js/autotrack/common.js
jmtaber129: The [comment describing getFiberNodeComponentHierarchyTraversal return type](https://github.com/heap/react-native-heap/pull/86/files#diff-7e0901a87b208cb50d8e49d8090b3770R49) might help answer your question.

Basically, the function used to return something like
```
AppComponent;[foo=bar];|MyScreen;|MyButton;|
```

Now it returns something like
```
[
  {
    elementName: 'AppComponent',
    fiberNode: <node ref>,
    propsString: '[foo=bar];',
  }, {
    elementName: 'MyScreen',
    ...
  },
  ...
]
```

I'll see about converting this to TS either in this PR or in a followup PR

js/autotrack/common.js
jmtaber129: Currently, the idea is:
* You start with some state of current HeapIgnore props
* You see an element that may or may not be a HeapIgnore component
* If it's a HeapIgnore component, you want to update the state of HeapIgnore props to take into account this HeapIgnore's props

`getUpdatedHeapIgnoreProps` would probably be more accurate

js/autotrack/touchables.js
jmtaber129: What is 'it' and 'that' in this context?

examples/TestDriver/src/pages/basics.js
jmtaber129: It's also worth noting that `HeapCaptureRestrictor` was merely in a comment on the original RFC, but once I started testing, I got a little frustrated about having to specify 3 different props when I really just wanted to ignore the target text of something.  I'll probably remove it before I merge this particular PR, just to keep this narrowed down to a subset of the feature scope in the RFC (although I'd be interested in revisiting the concept in the future)

timothyandrew: Could this be a method instead of a component, maybe?

The way I'm envisioning this, `Heap.withHeapIgnore` and `Heap.withRestrictedCapture` might set up `<HeapIgnore>` components with defaults and be the "API" for `<HeapIgnore>`, but users can use `<HeapIgnore>` directly for more granular control if they want to.

js/autotrack/common.js
```diff
@@ -57,9 +87,121 @@ const getFiberNodeComponentHierarchy = currNode => {
...
+    })
+    .join('');
+
+  return {
+    heapIgnoreProps: currentHeapIgnoreProps,
+    hierarchy: hierarchyStrings,
+  };
+};
+
+const BASE_HEAP_IGNORE_PROPS = {
```
timothyandrew: Can the HeapIgnore-specific functions be moved to `heapIgnore.js`?

js/autotrack/common.js
```diff
@@ -57,9 +87,121 @@ const getFiberNodeComponentHierarchy = currNode => {
...
+
+// Given an object array representing the component hierarchy from root to target component,
+// traverse through the element list to create the string representation of the hierarchy, taking
+// into account HeapIgnore specifications.
+// Returns an object containing:
+// * hierarchy - the string hierarchy representation of 'hierarchyArray'.
+// * heapIgnoreProps - the final set of HeapIgnore props that applies to the target component. Used
+//     to determine whether to include target text and/or ignore the entire interaction.
+const getHierarchyStringFromTraversal = hierarchyArray => {
+  let currentHeapIgnoreProps = {
```
timothyandrew: What do you think about encapsulating this in a (ideally Typescript) class? If we wanted to add a new property to this list in the future, we'd have to add it to 4 separate declarations in this file, which seems a bit error-prone.

examples/TestDriver/src/pages/basics.js
jmtaber129: This is just what was spec'd in the RFC.  I'm open to other names, but I'd want sign off from a few other people, too.

GrandPoohBear: That's fine.  I think my biggest complain was that it wasn't clear at first glance what the difference between ignoreInteraction and one of the things that came after it.

js/autotrack/touchables.js
GrandPoohBear: Sorry - yes that was vague.  I more just meant that it looked like a common action was being taken at the beginning of "autotrack" methods and I was wondering if there was a way to structure it so that _every_ autotrack method got that data by default.

examples/TestDriver/src/pages/basics.js
GrandPoohBear: I guess what's confusing is that I didn't realize you had a kind of an allowAll/disallowAll difference between the two - I just assumed that the HeapIgnore component would _of course_ ignore the interaction, so it doesn't even seem like it needs to be specified.  Possible I'm missing something, but if we go to the model where we only have a HeapIgnore component, does it make sense to keep this?

examples/TestDriver/src/pages/basics.js
```diff
@@ -14,26 +14,36 @@ import {
...
           testID="track1"
           title="Call Track1"
           onPress={() => Heap.track('pressInTestEvent1', {})}
         />
-        <Button
-          testID="track2"
-          title="Call Track2"
-          onPress={() => Heap.track('pressInTestEvent2', {})}
-        />
+        <HeapIgnore ignoreInteraction={false} ignoreTargetText={false}>
```
GrandPoohBear: Can you ignore the interaction (ignoreInteraction=true) but NOT ignore the targetText (targetText=false)... are there a limited number of configuration states that make sense?

js/autotrack/__tests__/heapIgnore.spec.js
```diff
@@ -0,0 +1,206 @@
...
+        .find({ testID: 'targetElement' })
+        .filter(Text);
+      const normalProps = getBaseComponentProps(normalComponent.instance());
+      expect(normalProps).toEqual({
+        touchableHierarchy:
+          'WrapperComponent;|Foo;|BarClass;|BarFunction;|HeapIgnore;|Text;[testID=targetElement];|',
+      });
+    });
+
+    it('No-op', () => {
```
GrandPoohBear: Can you give a more descriptive name for the test?  Is there a condition under which it should be a no-op?  I can imagine this test suddenly breaking months later because of cosmic rays, and having no idea what thing we should be looking for.

examples/TestDriver/src/pages/basics.js
jmtaber129: By default, HeapIgnore will ignore everything.  If you want it to _not_ ignore some part of an interaction (e.g. don't ignore the interaction or target text), you would specify that you should not ignore those aspects (e.g. `ignoreInteraction={false} ignoreTargetText={false}`).

Would this be more intuitive if all these were named such that they'd be false by default (e.g. `captureInteraction={true} captureTargetText={true}` for the above example)?  This was an alternative that was mentioned in the [RFC](https://paper.dropbox.com/doc/RFC-React-Native-HeapIgnore-API--AcbarA5TfEJTVOJaJH0XnXe4Ag-FospQLIad9QNQGQjEC4cD#:uid=622472213691434445521435&h2=Alternative-Ideas) that natty seemed to think would be a good idea, but no one else gave their input on it.

js/autotrack/touchables.js
jmtaber129: I don't see an easy way to do this without doing some weird reflection-y thing or adding an additional function call to every instrumentation site.  I think just having each autotrack method delegating to this for base props at the beginning of this method is sufficient.

examples/TestDriver/src/pages/basics.js
jmtaber129: This is a good question (which would've been great on the RFC 😉).  Certain config props do imply other config props (regardless of the value of the other prop).  For instance, you can't capture hierarchy props if you're not capturing the inner hierarchy (since the props you'd be capturing would go in the inner hierarchy).  Similarly, if you're not capturing the interaction, you can't capture _anything else_, since there's no interaction to attach these other things (target text, hierarchy, props etc).

js/autotrack/__tests__/heapIgnore.spec.js
jmtaber129: Ah yea, good catch.  Changed the `it` name and also added a TODO to the `HeapIgnore` component to add any other capture restrictions to it when we add them to `HeapIgnore` functionality (e.g. if we start capturing a new property, we'd probably add a prop to the HeapIgnore config, and we'd want to add it here to keep this a test for ignoring nothing).

jnatkins:
@jmtaber129 After some review of the API, I think we should make the following changes:

1. Refactor the property names to use the verb "allow" rather than "ignore". This changes the logic slightly, but will make it easier to incrementally loosen the restrictiveness of the HeapIgnore component. For example, 

```
<HeapIgnore>
  <Component>
</HeapIgnore>
```

Is the most restrictive setting, but then I can do something like

```
<HeapIgnore allowInnerHierarchy=true>
  <Component>
</HeapIgnore>
```

To be a little less restrictive, rather than having to add all props off the bat.

2. Add a convenience component for disabling target text capture. This is likely the second most common use case, so we should make it as easy as possible to do that. In order to facilitate that, I'd recommend adding a `<HeapIgnoreTargetText>` convenience component to shortcut the props on the HeapIgnore component.

js/autotrack/common.js
jmtaber129: yep!

js/autotrack/common.js
jmtaber129: This is definitely a possible refactor, but since this was simplified a bit with the removal of `HeapCaptureRestrictor`, I'd prefer to wait to pull this into its own class for now

jmtaber129:
From the implications of natty's comment, it seems like there may still be some confusion around how `ignoreXyz=...` actual worked, which I think says a lot about how bad this naming convention really is (basically `<HeapIgnore ignoreXyz=false>` is equivalent to `<HeapIgnore allowXyz=true>`) so I'm really glad we decided to move in this direction.

I was originally gonna try to merge as-is with the old prop config names, but since we got the updated config name concerns resolved so quickly, I decided I might as well include them in this PR instead of just changing them in a follow-up.

To reiterate, **this PR now uses the `allowXyz` config prop names instead of `ignoreXyz` prop names**

PR #87, opened 2019-05-14
[HEAP-9164] Convert common.js to Typescript

jmtaber129:
The logic implemented for HeapIgnore made it difficult to read and verify the types passed to and returned from new functions.  Convert common.js to typescript and add types.

This includes adding the `@types/react-reconciler` (but not `react-reconciler`) so we can get the `Fiber` typings.  `react-reconciler` is the package used by `react` for the fiber internals, so I believe we can depend on this with relatively low risk.

~Leaving this PR as a draft for now because this is based on https://github.com/heap/react-native-heap/pull/86 (not develop), but otherwise ready for review.~

~TODO: Rebase this onto develop once https://github.com/heap/react-native-heap/pull/86 is merged~


js/autotrack/common.ts
```diff
@@ -1,19 +1,59 @@
...
 // full interaction due to HeapIgnore.
 // Returns null if we're ignoring the full interaction due to HeapIgnore.
-export const getBaseComponentProps = componentThis => {
+export const getBaseComponentProps: (
+  component: Component
+) => AutotrackProps | null = componentThis => {
   // Get the hierarchy traversal from root to target component, then get the actual hierarchy from
   // the traversal representation.
-  const touchableHierarchyTraversal = getComponentHierarchyTraversal(
+  const touchableHierarchyTraversal: ComponentHierarchyTraversalElement[] = getComponentHierarchyTraversal(
```
GrandPoohBear: Do you have to do this explicitly here?  Shouldn't it get picked up by the return type of the func?

js/autotrack/common.ts
```diff
@@ -40,7 +80,9 @@ export const getBaseComponentProps = componentThis => {
   return autotrackProps;
 };
 
-const getComponentHierarchyTraversal = componentThis => {
+const getComponentHierarchyTraversal: (
+  comp: Component
+) => ComponentHierarchyTraversalElement[] = componentThis => {
   // :TODO: (jmtaber129): Remove this if/when we support pre-fiber React.
```
GrandPoohBear: I think it's probably safe to put aside this desire soon!

js/autotrack/common.ts
```diff
@@ -54,12 +96,10 @@ const getComponentHierarchyTraversal = componentThis => {
...
 // Traverse up the hierarchy from the current component up to the root, and return an array of
-// objects representing the component hierarchy from root to the current node. Each object element
-// contains:
-// * elementName - The name of the component
-// * fiberNode - The FiberNode reference for the component
-// * propsString - The string of props obtained from extractProps().
-const getFiberNodeComponentHierarchyTraversal = currNode => {
+// objects representing the component hierarchy from root to the current node.
+const getFiberNodeComponentHierarchyTraversal: (
+  currNode: FiberNode | null
```
GrandPoohBear: Hmm.... looking at line 104 below, it seems like the return type should be `FiberNode[]` ?  That way it _could_ be empty, right?

js/autotrack/common.ts
```diff
@@ -108,11 +148,16 @@ const getFiberNodeComponentHierarchyTraversal = currNode => {
...
+  hierarchyArray: ComponentHierarchyTraversalElement[]
+) => HierarchyResult = hierarchyArray => {
+  let currentHeapIgnoreProps: HeapIgnoreProps = _.mapValues(
+    BASE_HEAP_IGNORE_PROPS,
+    () => false
+  );
 
   // Map each hierarchy element to its string representation, considering HeapIgnore specs.
-  const hierarchyStrings = hierarchyArray
+  const hierarchyStrings: string = hierarchyArray
```
GrandPoohBear: Should this be plural?  At the end we get a single string out (of concatenated strings, but it's still just a single string out the other side...) . Perhaps just `hierarchyString`?

js/autotrack/common.ts
jmtaber129: Not sure what you mean.  `currNode.return` (the argument to the recursive call) is of type `FiberNode | null`, and the resultant hierarchy traversal of a given `FiberNode | null` is of type `ComponentHierarchyTraversalElement[]`

js/autotrack/common.ts
jmtaber129: This _could_ be implicit, but I personally prefer explicitly specifying the type here (since the type isn't immediately apparent from surrounding code).

I guess we don't have a style guide that talks about this, but to me, leaving off the type here would be akin to using the `auto` type from other languages, which I've been taught to avoid except with long and/or unnecessary type names.

js/autotrack/common.ts
jmtaber129: Yep, fixed

js/autotrack/common.ts
lukeautry: In the Heap project, we don't require explicit type annotation on assignments like this, but we do require explicit annotations on class property declarations and function return types. This is all enforced with `tslint`.

I think it's probably having a larger discussion around to develop some consensus. Personally, I prefer to use type inference in as many places as possible, but this is also means that code needs to be designed to be readable without accompanying type annotations. There is no hard and fast rule here, yet.

js/autotrack/common.ts
```diff
@@ -1,19 +1,60 @@
...
   // the traversal representation.
-  const touchableHierarchyTraversal = getComponentHierarchyTraversal(
+  const touchableHierarchyTraversal: ComponentHierarchyTraversalElement[] = getComponentHierarchyTraversal(
     componentThis
   );
-  const { hierarchy, heapIgnoreProps } = getHierarchyStringFromTraversal(
+  const {
+    hierarchy,
+    heapIgnoreProps,
+  }: HierarchyResult = getHierarchyStringFromTraversal(
```
lukeautry: The explicit annotation here seems pretty unnecessary since you already know what values you're pulling out.

js/autotrack/common.ts
jmtaber129: Okay, with Luke's explanation of where explicit type annotations are required, along with being reminded of [TS strict mode](https://mariusschulz.com/blog/typescript-2-3-the-strict-compiler-option), I'm realizing that an explicit here is _not_ how I've preferred to handle type annotations in the past.

Switched this to be an implicit typing

js/autotrack/common.ts
jmtaber129: fixed

PR #92, opened 2019-05-20
[HEAP-9052] Update documentation to cover auto-initialization.

timothyandrew:
I split iOS-specific and Android-specific instructions into separate files; doing this in one file was getting a bit unwieldy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,81 +23,10 @@ For autotrack, add the following plugins to your `.babelrc` or `babel.config.js`
 }
 ```
 
-### Manual setup - iOS with Cocoapods
+## Installation & Setup
 
-If you don't have a Podfile in the `ios` folder of your React Native project, follow this guide to create one: https://facebook.github.io/react-native/docs/integration-with-existing-apps#configuring-cocoapods-dependencies
-
-Add the following to your Podfile:
-
-```ruby
-pod "react-native-heap", path: "../node_modules/@heap/react-native-heap"
-```
-
-Then run:
-
-```bash
-pod install
-```
-
-You're done! :tada:
-
-### Manual setup - iOS without Cocoapods
-
-All terminal commands assume you are in the top-level directory of your React Native project.
-
-1. Locate the Xcode project for RNHeap: `open node_modules/@heap/react-native-heap/ios` . You should see a Finder window containing the `RNHeap.xcodeproj` file.
-1. Open the Xcode project for your native app, in the `ios` directory of your project.
-1. Drag the `RNHeap.xcodeproj` file from the Finder window into the "Libraries" group of your Xcode project.
-1. Select your project name in the Xcode project navigator. Select the "Build Phases" tab for your app's target.
-1. Expand the build phase called "Link Binary With Libraries". In the project navigator on the left, expand the `RNHeap.xcodeproj` entry, and find `libRNHeap.a` in the "Products" group. The icon for `libRNHeap.a` should look like a little building with columns.
-1. Drag the `libRNHeap.a` file into the list of other libraries to be linked.
-1. Expand the build phase called "Copy Bundle Resources". In the project navigator on the left, expand the `RNHeap.xcodeproj` entry, and find `HeapSettings.bundle`. Drag this into the list of bundle resources to be copied.
-
-**NOTE**: Using `react-native link` is not currently supported
-
-### Manual setup - Android
-
-In some cases, simply doing `react-native link` will correctly set up the tracker on Android.
-However, if you encounter any issues (either with the command itself or at compile-time/run-time),
-you should revert any changes made by `react-native link` and try the following steps, instead.
-
-Add the following to your `settings.gradle`:
-
-```groovy
-include ':react-native-heap'
-project(':react-native-heap').projectDir =
-    new File(rootProject.projectDir, '../node_modules/@heap/react-native-heap/android')
-```
-
-Then add the following to `app/build.gradle`:
-
-```groovy
-implementation project(':react-native-heap')
-```
-
-Import the package and add it to your app's MainApplication:
-
-```java
-
-@Override
-import com.heapanalytics.reactnative.RNHeapLibraryPackage;
-...
-protected List<ReactPackage> getPackages() {
-  return Arrays.<ReactPackage>asList(
-    new MainReactPackage(),
-    new RNHeapLibraryPackage()
-  );
-}
-```
-
-If you are seeing runtime warnings at startup mentioning `Heap: Could not find BuildConfig`, add the
-following line to the resources section of `res/values/strings.xml`:
-
-```xml
-<string name="com.heapanalytics.android.buildConfigPkgName">com.your_package_name</string>
-```
-
-and replace `com.your_package_name` with the package name from the manifest tag in `AndroidManifest.xml`.
+- [iOS](docs/ios-setup.md)
+- [Android](docs/android-setup.md)
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


docs/android-setup.md
```diff
@@ -0,0 +1,90 @@
...
+      "heapAppId": "11"
+    },
+    "prod": {
+      "heapAppId": "12",
+      "heapAutoInit": true
+    }
+  }
+  ```
+
+- Note that different application/environment IDs can be set for development and production. Heap can also be enabled/disabled on a per-environment basis. Values in `default` are used if a key is missing in either `dev` or `prod`.
```
jnatkins: What exactly triggers the use of dev or prod configurations?

docs/android-setup.md
```diff
@@ -0,0 +1,90 @@
...
+- In addition, find the `MainActivity.java` file, which is React Native's Android-specific entry point. This file is typically present in a subfolder of `android/app/src/main/java/`. Add an `onCreate` method containing:
+
+  ```java
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+      super.onCreate(savedInstanceState);
+      RNHeap.init(getApplicationContext(), "<app-id>");
+  }
+  ```
+- And the following imports:
```
jnatkins: Just a nit -- the imports should probably come first, since they're at the top of the file, and the onCreate function is later in the file.

docs/android-setup.md
```diff
@@ -0,0 +1,90 @@
...
+## Manual Initialization
+
+- If you'd like to skip auto-initialization, call `Heap.setAppId` with an application/environment ID.
+
+  ```javascript
+  import Heap from '@heap/react-native-heap';
+  Heap.setAppId('my-app-id');
+  ```
+
+- In addition, find the `MainActivity.java` file, which is React Native's Android-specific entry point. This file is typically present in a subfolder of `android/app/src/main/java/`. Add an `onCreate` method containing:
```
jnatkins: I'd just add "if it doesn't already exist" in reference to the onCreate method

docs/ios-setup.md
```diff
@@ -0,0 +1,98 @@
...
+      "heapAppId": "11"
+    },
+    "prod": {
+      "heapAppId": "12",
+      "heapAutoInit": true
+    }
+  }
+  ```
+
+- Note that different application/environment IDs can be set for development and production. Heap can also be enabled/disabled on a per-environment basis. Values in `default` are used if a key is missing in either `dev` or `prod`.
```
jnatkins: Same question from above

docs/ios-setup.md
```diff
@@ -0,0 +1,98 @@
...
+## Manual Initialization
+
+- If you'd like to skip auto-initialization, simply call `Heap.setAppId` with an application/environment ID.
+
+  ```javascript
+  import Heap from '@heap/react-native-heap';
+  Heap.setAppId('my-app-id');
+  ```
+
+- Note that manual initialization [on Android](./android-setup.md#manual-initialization) is slightly more involved.
```
jnatkins: I'd replace "is slightly more involved" with "requires additional steps"

docs/android-setup.md
timothyandrew: We use the `__DEV__` variable, which RN sets for non-release builds.

https://facebook.github.io/react-native/docs/javascript-environment#polyfills
https://facebook.github.io/react-native/docs/performance#profiling

jnatkins: I probably should have clarified that I was asking the question primarily because I think we should include that information in the docs

timothyandrew: Gotcha 👍

timothyandrew:
Thanks for the quick review @jnatkins! I'll send out the readme.io update soon.

README.md
```diff
@@ -23,89 +23,18 @@ For autotrack, add the following plugins to your `.babelrc` or `babel.config.js`
...
+- [Android](docs/android-setup.md)
 
 ## Usage
 
 ```js
 // Import Heap.
 import Heap from '@heap/react-native-heap';
 
-// Start Heap.
+// Start Heap if not using auto-initialization.
```
drob: If we expect auto-init to be the common case let's move this line out of the usage entirely and hide it away in the "manual initialization" section.

docs/ios-setup.md
```diff
@@ -0,0 +1,98 @@
...
+1. Open the Xcode project for your native app, in the `ios` directory of your project.
+1. Drag the `RNHeap.xcodeproj` file from the Finder window into the "Libraries" group of your Xcode project.
+1. Select your project name in the Xcode project navigator. Select the "Build Phases" tab for your app's target.
+1. Expand the build phase called "Link Binary With Libraries". In the project navigator on the left, expand the `RNHeap.xcodeproj` entry, and find `libRNHeap.a` in the "Products" group. The icon for `libRNHeap.a` should look like a little building with columns.
+1. Drag the `libRNHeap.a` file into the list of other libraries to be linked.
+1. Expand the build phase called "Copy Bundle Resources". In the project navigator on the left, expand the `RNHeap.xcodeproj` entry, and find `HeapSettings.bundle`. Drag this into the list of bundle resources to be copied.
+
+**NOTE**: Using `react-native link` is not currently supported
+
+# Initialization
```
drob: Nit: maybe we should call this "configuration", not "initialization". Under the hood it's the latter but from the user's point of view it's regular configuration. Wdyt?

docs/ios-setup.md
```diff
@@ -0,0 +1,98 @@
...
+      "heapAutoInit": true
+    }
+  }
+  ```
+
+- Note that different application/environment IDs can be set for development and production. Heap can also be enabled/disabled on a per-environment basis. Values in `default` are used if a key is missing in either `dev` or `prod`.
+
+## Manual Initialization
+
+- If you'd like to skip auto-initialization, simply call `Heap.setAppId` with an application/environment ID.
```
drob: I don't think users know what "auto-initialization" means. Why don't we say:
> If you'd like fine-grained control over when the Heap library initializes, call `Heap.setAppId` with an application/environment ID. (Most users won't need this.)

docs/ios-setup.md
timothyandrew: Works for me 👍

docs/ios-setup.md
timothyandrew: That's a lot better, thanks! 👍

docs/android-setup.md
```diff
@@ -0,0 +1,93 @@
+# Installation
+
+In some cases, simply doing `react-native link` will correctly set up the tracker on Android.
```
jmtaber129: > some cases

Is it still believed that `react-native link` often won't work?  I don't know if I've ever seen this _not_ work.

docs/android-setup.md
```diff
@@ -0,0 +1,93 @@
...
+If you are seeing runtime warnings at startup mentioning `Heap: Could not find BuildConfig`, add the
+following line to the resources section of `res/values/strings.xml`:
+
+  ```xml
+  <string name="com.heapanalytics.android.buildConfigPkgName">com.your_package_name</string>
+  ```
+
+and replace `com.your_package_name` with the package name from the manifest tag in `AndroidManifest.xml`.
+
+# Confguration
```
jmtaber129: A lot of this is shared with iOS.  I'd expect the most common use case for RN apps would be to install Heap for both iOS _and_ Android, not just one or the other, so this should probably be pulled out into a common set of instructions.

jmtaber129: Typo: "Confguration" -> "Configuration"

docs/android-setup.md
timothyandrew: This line is verbatim from the old Android section, I just moved it into this new file. Do you think it needs a change?

docs/android-setup.md
timothyandrew: There's enough platform-specific behavior in this section that managing it all in a single doc was unwieldy. I don't think it's unreasonable to expect users to follow two separate docs to integrate android and iOS, even if there's some repetition in there.

PR #93, opened 2019-05-20
[HEAP-9245] Add 'HeapIgnoreTargetText' convenience component

jmtaber129:
One HeapIgnore use case we expect to be somewhat common is to ignore a component subtree's target text and nothing else.  With the current `HeapIgnore` component, this would look like:
```
<HeapIgnore
  allowInteraction={true}
  allowInnerHierarchy={true}
  allowAllProps={true}
>
  <MySensitiveTextComponent />
</HeapIgnore>
```

To reduce the boilerplate for such a component, this PR adds a `HeapIgnoreTargetText` convenience component that is the equivalent of the above snippet, without the config:
```
<HeapIgnoreTargetText>
  <MySensitiveTextComponent />
</HeapIgnoreTargetText>
```

Note that one side effect of this particular implementation is that the presence of the `HeapIgnoreTargetText` component leads to a `HeapIgnoreTargetText;|HeapIgnore;|` in the resulting component hierarchy, rather than just `HeapIgnoreTargetText;|`.

### DIFF
--- a/js/autotrack/heapIgnore.js
+++ b/js/autotrack/heapIgnore.js
@@ -14,6 +14,7 @@ export const HeapIgnoreTargetText = props => {
       allowInteraction={true}
       allowInnerHierarchy={true}
       allowAllProps={true}
+      allowTargetText={false}
     >
       {props.children}
     </HeapIgnore>
--- a/js/autotrack/heapIgnore.js
+++ b/js/autotrack/heapIgnore.js
@@ -7,6 +7,19 @@ export const HeapIgnore = props => {
   return props.children;
 };
 
+// Convenience component for only ignoring target text.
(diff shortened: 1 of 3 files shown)


js/autotrack/heapIgnore.js
```diff
@@ -7,6 +7,19 @@ export const HeapIgnore = props => {
   return props.children;
 };
 
+// Convenience component for only ignoring target text.
+export const HeapIgnoreTargetText = props => {
+  return (
+    <HeapIgnore
+      allowInteraction={true}
+      allowInnerHierarchy={true}
+      allowAllProps={true}
```
GrandPoohBear: It would be great to be explicit that we have `allowTargetText={false}`.

PR #96, opened 2019-05-22
[HEAP-9116] HeapIgnore

jmtaber129:
Merge the `HeapIgnore` feature branch.  Each commit has been individually reviewed as a separate PR, so all that's needed is one final look before merging.

~Blocked by https://github.com/heap/react-native-heap/pull/95~. #95 has been merged


setup-tests.js
```diff
@@ -0,0 +1,18 @@
...
+configure({ adapter: new Adapter() });
+
+/**
+ * Ignore some expected warnings
+ * see: https://jestjs.io/docs/en/tutorial-react.html#snapshot-testing-with-mocks-enzyme-and-react-16
+ * see https://github.com/Root-App/react-native-mock-render/issues/6
+ */
+const originalConsoleError = console.error;
+console.error = (message) => {
+  if (message.startsWith('Warning:')) {
```
GrandPoohBear: Hmm... does this just squelch _anything_ that starts with `Warning:`?  If so, that seems a bit dangerous, even if it's just in tests!  Is there something more specific we can rely on, or is there a reason this is actually a lot more specific than I think?  Right now, it seems like an overly broad footgun...

jmtaber129: This is specifically `console.error()` calls that begin with `Warning:`, since this is apparently what React 16 outputs when mocked modules fail certain checks.  All `console.warn()` calls are unaffected, and any `console.error()`s that _don't_ start with `Warning:` are unaffected.

So we'll still see things like
```
    console.warn js/util/bailer.ts:12
      The Heap library has crashed with an exception. { name: undefined, message: undefined, stack: undefined }
    console.warn js/util/bailer.ts:12
      The Heap library has crashed with an exception. { name: undefined, message: undefined, stack: undefined }
```
and
```
console.warn foobar.ts:12
  Warning: something bad happened.
```
but won't have to deal with things like 
```
    console.error node_modules/react-dom/cjs/react-dom.development.js:506
      Warning: <Text /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.
          in Text (created by Text)
          in Text
          in BarFunction (created by Foo)
          in BarClass (created by Foo)
          in View (created by View)
          in View (created by Foo)
          in Foo (created by WrapperComponent)
          in WrapperComponent
    console.error node_modules/react-dom/cjs/react-dom.development.js:506
      Warning: The tag <Text> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.
          in Text (created by Text)
          in Text
          in BarFunction (created by Foo)
          in BarClass (created by Foo)
          in View (created by View)
          in View (created by Foo)
          in Foo (created by WrapperComponent)
          in WrapperComponent
```

This is specifically what was recommended in the Enzyme setup instructions

GrandPoohBear: Woah... weird...

PR #97, opened 2019-05-22
Fix android assertion on HeapIgnore

jmtaber129:
The `findEventInRedisRequests()` method used for assertions that an event did _not_ occur on iOS doesn't work for android, so add an equivalent spec for android that uses the `findAndroidEventInRedisRequests()` method instead.

### DIFF
--- a/e2e/heapIgnore.spec.js
+++ b/e2e/heapIgnore.spec.js
@@ -37,31 +37,10 @@ describe('HeapIgnore', () => {
     await rnTestUtil.pollForSentinel('HeapIgnore');
   });
 
-  it(':ios: should ignore the interaction', async () => {
+  it('should ignore the interaction', async () => {
     // Get all the events from redis, and assert that none of the requests match the ignored
     // interaction.
-    const { err, res } = await new Promise((resolve, reject) => {
-      testUtil.findEventInRedisRequests({}, (err, res) => {
-        resolve({ err, res });
-      });
-    });
-
-    assert.not.exist(err);
-    assert(res.length).not.equal(0);
-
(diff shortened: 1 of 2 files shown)


e2e/heapIgnore.spec.js
```diff
@@ -54,6 +54,23 @@ describe('HeapIgnore', () => {
     assert(JSON.stringify(res)).not.match(/Foo;\|/)
   });
 
+  it(':android: should ignore the interaction', async () => {
+    // Get all the events from redis, and assert that none of the requests match the ignored
+    // interaction.
+    const { err, res } = await new Promise((resolve, reject) => {
+      testUtil.findAndroidEventInRedisRequests({}, (err, res) => {
```
GrandPoohBear: What do you think about doing this switch within a separate fn?  So make something like `findEvent({})` that does the appropriate switch based on the running platform?  That way we don't have to duplicate the rest of the test.  I _believe_ you get access to that as a global while detox is running...

e2e/heapIgnore.spec.js
```diff
@@ -40,18 +40,14 @@ describe('HeapIgnore', () => {
   it('should ignore the interaction', async () => {
     // Get all the events from redis, and assert that none of the requests match the ignored
     // interaction.
-    const { err, res } = await new Promise((resolve, reject) => {
-      testUtil.findEventInRedisRequests({}, (err, res) => {
-        resolve({ err, res });
-      });
-    });
+    const { err, res } = await rnTestUtil.findAllEvents();
```
GrandPoohBear: This is nice dude... good work!

PR #102, opened 2019-06-12
[HEAP-9594] Add RN Elements builtin prop config

jmtaber129:
Built-in props for the React Native Elements component library.

### DIFF
--- a/js/propExtractorConfig.ts
+++ b/js/propExtractorConfig.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 
 import { PropExtractorConfig } from './util/extractProps';
+import { REACT_NATIVE_ELEMENTS_CONFIG } from './libraryPropConfigs/rnElementsPropConfig';
 
 const BASE_CONFIG: PropExtractorConfig = {
   '*': {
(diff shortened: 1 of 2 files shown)


js/propExtractorConfig.ts
```diff
@@ -1,5 +1,76 @@
 import { PropExtractorConfig } from './util/extractProps';
 
+const REACT_NATIVE_ELEMENTS_CONFIG: PropExtractorConfig = {
```
GrandPoohBear: This looks great!  I think I might have a slight preference to keep this in its own file and import it into here?  Maybe create a directory with default configs?

js/propExtractorConfig.ts
```diff
@@ -1,5 +1,76 @@
...
+      'showEditButton',
+      'title',
+    ],
+    exclude: [],
+  },
+  Badge: {
+    include: ['status', 'value'],
+    exclude: [],
+  },
+  Button: {
```
GrandPoohBear: I think that the way you have this now, this `Button` entry will override the default RN `Button` entry.  What we really need to do is do a deep merge instead of the object splat, I think...

GrandPoohBear: Alternately, we could just include `title` in here, though I like being able to keep them separate...

js/propExtractorConfig.ts
drob: +1, think this is smart.

PR #105, opened 2019-06-18
[HEAP-9279] Instrument ScrollView paging

jmtaber129:
Instrument paging on `ScrollView`s when `pagingEnabled`.  I.e., autocapture when a user swipes on a `ScrollView` that performs paging (i.e. snapping to intervals equal to the width of the view).

Event specs:
* Name: `scrollViewPage`,
* Props:
  * `touchableHierarchy`: the hierarchy (we should really change this prop to just be `hierarchy` across the board)
  * `pageIndex`: The calculated index of the page scrolled to (determined from `offset / view width`)

Note that `targetText` is not included in event props, since the target text of a scrollview is the _entire_ text contents of the scrollview, which is a _lot_ of text (Open Q: should we just start truncating target text across the board, and then include target text in this event type?) 

A few caveats:
* Configurations that implement the effect of paging via `snapTo*` props are *not* captured
* Events still fire even when the `pageIndex` is **not** changed.  E.g. if a user starts scrolling, but the view ends up snapping to the last position, an event still fires.  **TODO**: Prioritize changing this to only fire when the new `pageIndex` is different from the previous `pageIndex`.


instrumentor/src/index.js
```diff
@@ -73,6 +73,43 @@ const instrumentTouchables = path => {
...
 
+const instrumentScrollView = path => {
+  if (path.node.key.name === 'onMomentumScrollEnd') {
+    // Find the parent 'props' declaration.
+    const propsParent = path.findParent(path => {
+      return path.isVariableDeclarator() && path.node.id.name === 'props';
+    });
+
+    if (!propsParent) {
+      // console.log('no props parent');
```
GrandPoohBear: nit - cruft?

e2e/basics.spec.js
```diff
@@ -24,6 +24,12 @@ const doTestActions = async () => {
   await element(by.id('track3')).tap();
   await element(by.id('clearProps')).tap();
   await element(by.id('track4')).tap();
+
+  // :HACK: Break up long URL.
```
GrandPoohBear: yaaaarrrrrrgggghhhh!!

instrumentor/src/index.js
```diff
@@ -81,7 +81,6 @@ const instrumentScrollView = path => {
     });
 
     if (!propsParent) {
-      // console.log('no props parent');
```
GrandPoohBear: lol, nevermind... danger of going commit by commit

PR #107, opened 2019-06-27
Pin detox version

jmtaber129:
There seems to be a bug in later versions of detox that are causing android tests to hang when I run locally.  Pin detox to a specific version instead of using `^`.

### DIFF
--- a/examples/TestDriver/package.json
+++ b/examples/TestDriver/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "babel-jest": "23.6.0",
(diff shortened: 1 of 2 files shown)


examples/TestDriver/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "babel-jest": "23.6.0",
-    "detox": "^12.1.1",
+    "detox": "12.1.1",
```
kamalmarhubi: I'd say leave a comment in here, but lol json :/

PR #108, opened 2019-06-27
Fix scrollview tests on android

jmtaber129:


js/autotrack/scrollViews.js
```diff
@@ -29,7 +29,9 @@ export const autotrackScrollView = track => (
     event.nativeEvent.contentOffset.x /
     event.nativeEvent.layoutMeasurement.width;
 
-  autotrackProps.pageIndex = pageIndex;
+  // Integer props on android seem to come into Heap as decimals (regardless of whether the number is actually a float), while integer props
+  // on iOS come into Heap as integers. To keep this prop consistent across platforms, send it as a string.
+  autotrackProps.pageIndex = `${Math.round(pageIndex)}`;
```
kamalmarhubi: do you mean "come into Heap" in device or in server? is there a bigger change we should be making somewhere?

jmtaber129: It's not clear if this is happening when it goes through android or collectors or what.  The number goes through the RN->Android bridge as an integer, and shows up in the UI as a decimal.

PR #114, opened 2019-07-16
[HEAP-9951] Capture screen name props for autocaptured events

jmtaber129:
This adds `path` and `screenName` props to all autocaptured events (e.g. `touchableHandlePress`, etc.), where `path` is the full path of the current screen route (e.g. `Nav::MainStack::Base`), and `screenName` is the individual name of the current screen route (e.g. `Base`).

This does *not* include these props on custom tracked events.

### DIFF
--- a/js/autotrack/reactNavigation.js
+++ b/js/autotrack/reactNavigation.js
@@ -48,6 +48,7 @@ export const withReactNavigationAutotrack = track => AppContainer => {
         <AppContainer
           ref={bailOnError(navigatorRef => {
             this.setRef(forwardedRef, navigatorRef);
+            // Update the NavigationUtil's nav reference to the updated ref.
             NavigationUtil.setNavigationRef(navigatorRef);
             // Only update the 'topLevelNavigator' if the new nav ref is different and non-null.
             if (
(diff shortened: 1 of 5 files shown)


js/autotrack/reactNavigation.js
```diff
@@ -48,6 +48,7 @@ export const withReactNavigationAutotrack = track => AppContainer => {
         <AppContainer
           ref={bailOnError(navigatorRef => {
             this.setRef(forwardedRef, navigatorRef);
+            NavigationUtil.setNavigationRef(navigatorRef);
```
GrandPoohBear: Probably worth an explanatory comment!

js/util/navigationUtil.ts
```diff
@@ -1,14 +1,48 @@
...
+
+    if (routeProps) {
+      return routeProps;
+    }
+
+    return null;
+  }
+
+  // :TODO: (jmtaber129): Add type for navigationState.
+  static getActiveRouteProps(
```
GrandPoohBear: Perhaps return an array of screen strings, so that the last one is the current screen, and the rest can be glommed together to paths

PR #120, opened 2019-08-05
[HEAP-9795] Reset Identity

jmtaber129:
Add the `resetIdentity` method to the bridge + add tests.

### DIFF
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -173,6 +173,9 @@ describe('Basic React Native and Interaction Support', () => {
       assert.not.exist(err2);
       assert(res1.length).not.equal(0);
       assert(res2.length).not.equal(0);
+
+      // Assert that there is no identity post-resetIdentity, and that the post-resetIdentity user ID is different from the
+      // pre-resetIdentity user ID.
       assert.not.exist(res2[0]['i']);
       assert(res1[0]['u']).not.equal(res2[0]['u']);
     });
@@ -314,6 +317,9 @@ describe('Basic React Native and Interaction Support', () => {
       assert.not.exist(err2);
       assert(res1.length).not.equal(0);
       assert(res2.length).not.equal(0);
+
+      // Assert that there is no identity post-resetIdentity, and that the post-resetIdentity user ID is different from the
+      // pre-resetIdentity user ID.
       assert.not.exist(res2[0]['user']['identity']);
       assert(res1[0]['user']['id']).not.equal(res2[0]['user']['id']);
     });
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -56,6 +56,13 @@ const doTestActions = async () => {
 
   await element(by.id('scrollView')).swipe('left');
 
+  // :HACK: Break up long URL.
+  // :TODO: Remove once pixel endpoint is handling larger events again.
+  console.log('Waiting 15s to flush iOS events.');
+  await new Promise(resolve => setTimeout(resolve, 15000));
+
+  await element(by.id('resetIdentity')).tap();
+
   await element(by.id('basicsSentinel')).tap();
 };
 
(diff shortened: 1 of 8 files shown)


daviscabral:
Hey guys, not sure if this is to fix the ScrollView error, but now it does not trigger the error anymore but crashes the Heap lib with an exception causing just an warning (what I believe it's much better for now). 

Screenshot 2019-08-06 10 32 25

e2e/basics.spec.js
```diff
@@ -146,6 +153,23 @@ describe('Basic React Native and Interaction Support', () => {
       assert(res.length).equal(1);
       assert(res[0]['i']).equal('foobar');
     });
+
+    it('should reset identity', async () => {
+      const { err1, err2, res1, res2 } = await new Promise(resolve => {
+        testUtil.findEventInRedisRequests({ t:  'pressInTestEvent2'}, (err1, res1) => {
```
GrandPoohBear: It's a little hard to follow what's happening here... Certainly, there's a lower threshold for readability, since this is test code, but perhaps factoring out a `findEventAfterSentinel` method or something would make it clearer?

e2e/basics.spec.js
```diff
@@ -146,6 +153,23 @@ describe('Basic React Native and Interaction Support', () => {
...
+            resolve({ err1, err2, res1, res2 });
+          });
+        });
+      });
+
+      assert.not.exist(err1);
+      assert.not.exist(err2);
+      assert(res1.length).not.equal(0);
+      assert(res2.length).not.equal(0);
+      assert.not.exist(res2[0]['i']);
```
GrandPoohBear: Would you mind adding a comment as to what these keys are, and what you're actually checking here?

e2e/basics.spec.js
```diff
@@ -269,7 +293,7 @@ describe('Basic React Native and Interaction Support', () => {
   describe('Autotrack', () => {
     it("should autotrack 'TouchableOpacity's", async () => {
       const expectedHierarchy =
-        'AppContainer;|App;|Provider;|HeapNavigationWrapper;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;[key=Basics];|SceneView;|Connect(BasicsPage);|BasicsPage;|TouchableOpacity;[testID=touchableOpacityText];|';
+        'AppContainer;|App;|Provider;|HeapNavigationWrapper;|NavigationContainer;|Navigator;|NavigationView;|TabNavigationView;|ScreenContainer;|ResourceSavingScene;[key=Basics];|SceneView;|Connect(BasicsPage);|BasicsPage;|ScrollView;|TouchableOpacity;[testID=touchableOpacityText];|';
```
GrandPoohBear: Maybe it's time to factor a large prefix of this string into a constant and share it between the tests?  Sucks to have to modify _n_ tests for every minor UI change in TestDriver...

GrandPoohBear:
@daviscabral - this particular change is _not_ to address that issue, but it's on our radar, and we're actively working on getting a fix for that.  We'll be sure to flag it for you when it's fixed!

daviscabral:
@GrandPoohBear no worries, thanks, waiting on that to proceed with complete integration on product here.

e2e/basics.spec.js
jmtaber129: Could you elaborate on what you mean by `findEventAfterSentinel` and what you'd like pulled out here?

e2e/basics.spec.js
jmtaber129: Yea, I agree, although this should probably happen in a follow-up PR

e2e/basics.spec.js
GrandPoohBear: I just mean that you have to jump through a couple of hoops to get what you want here, which makes this section particular hard to grok for the uninitiated.  It seemed to me like what you were really saying was "find this event after we're sure we've seen the sentinel".  Pulling that out into a named function might make this a lot easier to read for someone who's new to the project.

jmtaber129: All of these assertions happen after the sentinel anyway (since the `before` hook blocks until we see the sentinel).  I assume you mean before+after the `resetIdentity` call?

I don't think it's worth pulling this into a separate function.  Added comments for clarifying what this is doing

Issue #119, opened 2019-08-01
ScrollView Instrumentation breaks the component


cesarp:
When I enabled auto tracking in my application with the babel plugin using @heap/react-native-heap 0.6.0 and react-native 0.60.4 the ScrollView component crashes when the  `onMomentumScrollEnd` callback gets executed.

It throws an exception saying `_this is not defined` which by the way is correct, it isn't defined on that scope where it is trying to get access to it.

gist with the generated code that crashes the app:
https://gist.github.com/cesarp/92ed29d7035a0a9551fc7893aa3f1fd3#file-scrollview-js-L464

My system configuration:

```
System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
    Memory: 706.92 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.16.0 - ~/.nvm/versions/node/v8.16.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v8.16.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 23, 24, 25, 26, 27, 28
      Build Tools: 23.0.1, 25.0.1, 25.0.2, 26.0.2, 26.0.3, 27.0.3, 28.0.1, 28.0.3
      System Images: android-26 | Google APIs Intel x86 Atom_64, android-27 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.4 => 0.60.4
```

daviscabral:
Same here.

jnatkins:
Hi @cesarp and @daviscabral, wanted to give a heads up that our eng team is looking into this, and will update with more details.

Thanks!

jnatkins:
[...]

PR #121, opened 2019-08-07
[HEAP-10061] Fix scrollview `this` issue

jmtaber129:
~Leaving this as a draft until I confirm the most recent changes work on both RN 0.57.5 and RN 0.60.4, but otherwise ready for review~.  I haven't been able to get RN 0.60.4 to build Android on the test app, but I got an upgraded iOS build to a state in which unit+e2e tests pass on RN 0.60.4.  Because this was an issue with the babel plugin (not native code), I think it's reasonable to go ahead and merge.

On newer versions of react native (after the `ScrollView` component was switch from using a mixin and `createReactClass` to extending `React.Component`), the referenced `this` is incorrect, and causes errors upon `onMomentumScrollEnd()`.

Change the `ScrollView` instrumentation to use `t.thisExpression()` instead of `t.identifier('_this')`.
TODO: look into using `t.thisExpression()` for other instrumentations (this doesn't seem to work for these out-of-the-box).

Fixes https://github.com/heap/react-native-heap/issues/119

### DIFF
--- a/instrumentor/src/index.js
+++ b/instrumentor/src/index.js
@@ -65,7 +65,7 @@ const instrumentTouchables = path => {
 
     const replacementFunc = getOriginalFunctionReplacement(
       path.node.value, // originalFunctionExpression
-      'this', // thisIdentifier
+      t.identifier('this'), // thisIdentifier
       'autotrackPress', // autotrackMethodName
       path.node.key.name // eventType
     );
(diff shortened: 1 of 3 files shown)


jmtaber129:
Tests already exist for `ScrollView`s; the issue here wasn't lack of existing tests, but rather just not testing on newer versions of RN.

PR #124, opened 2019-08-09
[HEAP-9669] Ref-forwarding on HeapIgnore HOC

jmtaber129:
Forward references passed to the HeapIgnore HOC.

### DIFF
--- a/js/autotrack/__tests__/heapIgnore.spec.js
+++ b/js/autotrack/__tests__/heapIgnore.spec.js
@@ -193,7 +193,10 @@ describe('Common autotrack utils', () => {
     });
 
     it('Works as an HOC', () => {
-      const IgnoredText = withHeapIgnore(Text, { allowInteraction: true, allowInnerHierarchy: true });
+      const IgnoredText = withHeapIgnore(Text, {
+        allowInteraction: true,
+        allowInnerHierarchy: true,
+      });
       const wrapper = mount(
         <Foo>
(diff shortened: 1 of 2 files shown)


js/autotrack/__tests__/heapIgnore.spec.js
```diff
@@ -193,7 +193,7 @@ describe('Common autotrack utils', () => {
     });
 
     it('Works as an HOC', () => {
-      const IgnoredText = withHeapIgnore(Text, { allowInteraction: true });
+      const IgnoredText = withHeapIgnore(Text, { allowInteraction: true, allowInnerHierarchy: true });
```
jmtaber129: Updating this to test that `HeapIgnore` children make it through the HOC.

js/autotrack/heapIgnore.js
```diff
@@ -22,17 +22,25 @@ export const HeapIgnoreTargetText = props => {
...
+          <IgnoredComponent ref={forwardedRef} {...rest}>
             {this.props.children}
           </IgnoredComponent>
         </HeapIgnore>
       );
     }
   };
+
+  // :TODO: (jmtaber129): Change this to 'withHeapIgnore(<IgnoredComponent name>).
+  WithHeapIgnore.displayName = '_class';
```
jmtaber129: I'm keeping the display name as-is to avoid potentially-breaking changes to hierarchies

PR #125, opened 2019-08-14
Upgrade iOS lib to 6.2.1, bump RN lib version to 0.7.1

GrandPoohBear:
The libHeap v6.2.0 that had previously been included with this package had a nasty bug where user IDs would not be reused across sessions (more or less).  This PR updates the embedded lib to 6.2.1, and bumps the version of this lib to 0.7.1

### DIFF
--- a/examples/TestDriver/ios/TestDriver.xcodeproj/project.pbxproj
+++ b/examples/TestDriver/ios/TestDriver.xcodeproj/project.pbxproj
@@ -358,7 +358,7 @@
 			inputFileListPaths = (
 			);
 			inputPaths = (
-				"${SRCROOT}/Pods/Target Support Files/Pods-TestDriver/Pods-TestDriver-resources.sh",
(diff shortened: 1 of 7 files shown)


examples/TestDriver/ios/TestDriver.xcodeproj/project.pbxproj
```diff
@@ -358,7 +358,7 @@
 			inputFileListPaths = (
 			);
 			inputPaths = (
-				"${SRCROOT}/Pods/Target Support Files/Pods-TestDriver/Pods-TestDriver-resources.sh",
+				"${PODS_ROOT}/Target Support Files/Pods-TestDriver/Pods-TestDriver-resources.sh",
```
GrandPoohBear: Looks like this is a side effect of using Cocoapods 1.6.1 instead of 1.5.3.  The change is entirely innocuous, but happy to pull it out of this PR if there any objections.

PR #129, opened 2019-08-16
[HEAP-10401] Implement TextInput instrumentation

jmtaber129:
Implement per the [RFC](https://paper.dropbox.com/doc/miniRFC-React-Native-TextInput-Capture--AidTkeTtTn8Gk_7KQ7qnPeJDAg-0877UPqDFpKmZX67QPcp1)

### DIFF
--- a/js/autotrack/textInput.js
+++ b/js/autotrack/textInput.js
@@ -1,0 +1,37 @@
+import { getBaseComponentProps } from './common';
+import * as _ from 'lodash';
+
+export const autocaptureTextInputChange = track => (
+  eventType,
+  componentThis,
+  event
+) => {
+  // Attach a debounce function to the TextInput component instance if one's not already attached.
+  if (!componentThis.__heap__debounceTextChange) {
+    componentThis.__heap__debounceTextChange = _.debounce(
+      debouncedAutocaptureTextInputChange(track),
+      1000
+    );
+  }
+
+  componentThis.__heap__debounceTextChange(eventType, componentThis, event);
+};
+
+const debouncedAutocaptureTextInputChange = track => (
+  eventType,
+  componentThis,
+  event
+) => {
+  const autotrackProps = getBaseComponentProps(componentThis);
+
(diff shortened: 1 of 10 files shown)


js/autotrack/textInput.js
```diff
@@ -0,0 +1,37 @@
...
+      debouncedAutocaptureTextInputChange(track),
+      1000
+    );
+  }
+
+  componentThis.__heap__debounceTextChange(eventType, componentThis, event);
+};
+
+const debouncedAutocaptureTextInputChange = track => (
+  eventType,
```
GrandPoohBear: It might be helpful to add a comment to link to the event types that we're expecting (could just be a link to RN source on github or a doc if they have it)

js/autotrack/textInput.js
```diff
@@ -0,0 +1,37 @@
...
+  event
+) => {
+  const autotrackProps = getBaseComponentProps(componentThis);
+
+  if (!autotrackProps) {
+    // We're not capturing this interaction.
+    return;
+  }
+
+  if (componentThis.props.placeholder) {
```
GrandPoohBear: Do we need to check that `componentThis` is real here?  Would it ever come through as null?  (might not be the worst thing to check regardless...)

js/autotrack/textInput.js
jmtaber129: In this case, the event type is included as a string literal that we specify in the instrumentation code: https://github.com/heap/react-native-heap/pull/129/files#diff-df6ad962b9c024cab3a8be043f101c81R142

js/autotrack/textInput.js
jmtaber129: `componentThis` should never be `null`/`undefined`, but if it is, an error would happen in the `getBaseComponentProps()` call.  We should probably do some sort of invariant assertion within all instrumentation hook methods for these arguments

GrandPoohBear: 👍

PR #130, opened 2019-08-16
enable build with java 8 components

millarm:
Enables build with other Java8 components e.g. https://github.com/wix/Detox

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,7 +50,10 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 26)
-
+    compileOptions {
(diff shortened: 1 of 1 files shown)


kamalmarhubi:
Hi @millarm, thanks for the PR! I just wanted to check: is it currently not possible to use the latest version of detox? My assumption is that our library using the default of 1.7 shouldn't affect ability to use dependencies compiled differently.

We'll be happy to merge this, but we just want to fully understand the motivation. :-)

millarm:
No, detox is built using 1.8 - so at the linking stage it seems the built tool can't translate everything unless it's all built with 1.8.   With this change we seem to work fine

Matt

PR #131, opened 2019-08-16
Grab bag of test improvements

jmtaber129:
A few small test improvements (best reviewed commit-by-commit):
* Refactor page-level hierarchies into constant
* Only wait for iOS to flush on iOS tests (reduces android test run times by minutes)
* Fix nesting for Basics (previously, a failed `before` for autocapture tests wouldn't prevent autocapture tests from running)
* Pull redis `flushall()` into promisified util
* Reduce android flakes caused by the keyboard being in the way when a button should be tapped

### DIFF
--- a/e2e/rnTestUtilities.js
+++ b/e2e/rnTestUtilities.js
@@ -7,6 +7,15 @@ testUtil = require('../../heap/test/util');
 
 const HEAP_ENV_ID = '2084764307';
 
+const waitIfIos = async () => {
+  if (device.getPlatform() === 'ios') {
+    // :HACK: Break up long URL.
+    // :TODO: Remove once pixel endpoint is handling larger events again.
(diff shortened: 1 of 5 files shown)


e2e/rnTestUtilities.js
```diff
@@ -1,12 +1,26 @@
...
 const _ = require('lodash');
+const nodeUtil = require('util');
 
 db = require('../../heap/back/db');
 testUtil = require('../../heap/test/util');
 
 const HEAP_ENV_ID = '2084764307';
 
+const waitIfIos = async () => {
+  if (device.getPlatform() === 'ios') {
```
GrandPoohBear: Be nice to just get rid of this!  Have you looked into the server arg we need to start the collector with?

jmtaber129: I have, but I haven't been able to get it to work as expected :(

PR #132, opened 2019-08-22
[HEAP-10168] Don't fail build when dev or prod isn't defined in heap.config.json

jmtaber129:
Currently, if either the `dev` or `prod` object in `heap.config.json` is not present, the build will crash. For example, if `heap.config.json` is 
```
{
    "default": {
        "heapAutoInit": true,
        "heapAppId": "11"
    },
    "prod": {
        "heapAppId": "12"
    }
}
```
the build will fail.  Currently, you'd need to add a `dev` object, like
```
{
    "default": {
        "heapAutoInit": true,
        "heapAppId": "11"
    },
    dev: {},
    "prod": {
        "heapAppId": "12"
    }
}
```
in order to fix this.  Instead, just accept configs missing these fields.


ios/HeapSettings.bundle/generate_settings_common.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  settings = template
+    .gsub('__HEAP_APP_ID_DEV__', app_id_dev || '')
+    .gsub('__HEAP_APP_ID_PROD__', app_id_prod || '')
+
+  IO.write output_path, settings
+end
+
+def get_heap_settings_from_json(json)
+  dev, prod, default = json.values_at('dev', 'prod', 'default')
```
GrandPoohBear: OK, been spinning on this a bunch because I'd like to figure out a way to do this that's not too messy, but the way it's structured makes it complicated.  It's not good enough to say something like `app_id_dev = (dev and dev['heapAutoInit'] and dev['heapAppId']) or (default ...` because that `heapAutoInit` key might legitimately be false... which is weird... it seems like we shouldn't have even given you the option if the file was there at all, but what can you do? (could probably easily deprecate or change without pissing off too many people I bet...)

I'm thinking the easiest thing to do here is just to check if they're `nil` right off the top and then substitute empty dictionaries if they are.  Something like:
```suggestion
dev = json.values_at('dev') or {}
prod = json.values_at('prod') or {}
default = json.values_at('default') or {}
```

... then lines 27 and 28 go back to their original formulation.  What do you think?

ios/HeapSettings.bundle/generate_settings_non_cocoapods.rb
```diff
@@ -3,48 +3,29 @@
...
-    auto_init_dev = default['heapAutoInit'] if auto_init_dev.nil?
-    auto_init_dev = true if auto_init_dev.nil?
-
-    auto_init_prod = prod['heapAutoInit']
-    auto_init_prod = default['heapAutoInit'] if auto_init_prod.nil?
-    auto_init_prod = true if auto_init_prod.nil?
 
-    app_id_dev = dev['heapAppId'] || default['heapAppId'] if auto_init_dev
-    app_id_prod = prod['heapAppId'] || default['heapAppId'] if auto_init_prod
+    app_id_dev, app_id_prod = get_heap_settings_from_json(json)
```
GrandPoohBear: Yeah, that duplication is nasty.  Nice one.

ios/HeapSettings.bundle/generate_settings_common.rb
jmtaber129: Ah, I spent the better part of an hour trying to figure out how best to simplify this logic, but I hadn't even considered this.  This seems like the way to go

ios/HeapSettings.bundle/generate_settings
```diff
@@ -3,53 +3,34 @@
 require "json"
 require 'fileutils'
```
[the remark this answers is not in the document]
jmtaber129: Yep.  Fixed

ios/HeapSettings.bundle/generate_settings_common.rb
```diff
@@ -0,0 +1,30 @@
...
+  output_path = File.join dir, 'HeapSettings.plist'
+
+  settings = template
+    .gsub('__HEAP_APP_ID_DEV__', app_id_dev || '')
+    .gsub('__HEAP_APP_ID_PROD__', app_id_prod || '')
+
+  IO.write output_path, settings
+end
+
+def get_heap_settings_from_json(json)
```
GrandPoohBear: Well, damn... this is just beautiful now...

jmtaber129: 🚫code duplication

PR #136, opened 2019-08-29
Add basic web support

tberman:
Pretty basic first pass to support react-native-web via a small wrapper around the native module backstopping it with the heap js library.

autotracking presses doesn't seem to work, which I may dig into, but this seemed like a good place to start and get some thoughts.

### DIFF
--- a/js/Heap.js
+++ b/js/Heap.js
@@ -1,6 +1,5 @@
 // Libraries
 import React from 'react';
-import { NativeModules } from 'react-native';
 
 import {
   HeapIgnore,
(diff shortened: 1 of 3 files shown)


jmtaber129:
Hi Todd, thanks for the PR.  We are internally discussing support for React Native Web, and will decide how to proceed with your proposed changes early the week of 9/16.

NicholasMiller:
Hi @jmtaber129! Any update on the status of this PR?

jmtaber129:
@jnatkins followed up with the author via email a couple months ago, but we never actually responded here.  Sorry for the delay!  The gist is:

We took a look at this, and while the patch looks pretty good from our perspective, we're not sure that we're quite ready to support this for our entire customer-base yet, since it'll have broader implications in terms of how we continue to support our React Native integration.

We're tracking this in our roadmapping tool, in case other folks start expressing interest, but we're unlikely to merge this at the moment.  

That said, if this is working for you, great!  You're more than welcome to install this package directly off of your fork (e.g. via something like `npm i getmargin/react-native-heap#add_basic_web_support`, and rebase when we put out new releases. If you run into one-off issues with this, we're happy to work to try to resolve them, but want to set appropriate expectations for how we're going to be able to support this for the time being.

Closing this for now.

PR #138, opened 2019-09-06
[HEAP-9338] Add PR template

jmtaber129:
We've regularly forgotten to do things like update the changelog when adding a new feature (vs updating the changelog when we're about to do a release), so add a basic PR template including tasks like "update the changelog"

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,7 @@
+## Description
+
+## Test Plan
+
+## Checklist
+- [ ] Detox tests pass (only Heap employees are able run these)
+- [ ] If this is a bugfix/feature, the changelog has been updated
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,12 @@
 ## Description
+What does this PR add/fix/change?
+Is there anything reviewers should pay special attention to?
+Are there any breaking changes?
 
 ## Test Plan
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+## Description
+
```
GrandPoohBear: Might be good to put some prompts in here, maybe something like: 
What does this PR add/fix/remove?
What should reviewers pay special attention to?
Any breaking changes?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
...
+## Test Plan
+
+## Checklist
+- [ ] Detox tests pass (only Heap employees are able run these)
```
GrandPoohBear: I'm not as familiar with how those checklists work... If someone from outside the org submits a PR, can someone on our team then check those boxes?

PULL_REQUEST_TEMPLATE.md
jmtaber129: Done

PULL_REQUEST_TEMPLATE.md
jmtaber129: These are checked by placing an `X` in the box, e.g. 
```
- [X] XYZ thing
```
yields
- [X] XYZ thing

I think it would be reasonable for a Heap employee to edit the PR description to check the detox test box for PRs from outside the org.

PR #142, opened 2019-09-17
[HEAP-10464] Use correct conventions for HOC display name

jmtaber129:
## Description
Use `displayName` conventions for higher-order components (described [here](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging)).

This will break any event definitions relying on the component naming of `HeapNavigationWrapper` (for `withReactNavigationAutotrack`) or `_class` (for `withHeapIgnore`).  These aren't semantically meaningful, and likely aren't used for event def.

## Test Plan
* Updated all detox test to use new display names in hierarchies.
* Updated HeapIgnore unit test to use new display names in hierarchies.
* Added detox test case for HeapIgnore HOC display name.

## Checklist

### DIFF
--- a/js/autotrack/heapIgnore.js
+++ b/js/autotrack/heapIgnore.js
@@ -1,6 +1,8 @@
 import * as _ from 'lodash';
 import React from 'react';
 
+import { getComponentDisplayName } from '../util/hocUtil';
+
 export const HeapIgnore = props => {
   // The only purpose of HeapIgnore is to tell the us which props apply to the subtree when we're
   // traversing, so we just need to render it's children here.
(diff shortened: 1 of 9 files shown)


js/autotrack/reactNavigation.js
```diff
@@ -84,6 +85,10 @@ export const withReactNavigationAutotrack = track => AppContainer => {
     }
   }
 
+  HeapNavigationWrapper.displayName = `withReactNavigationAutotrack(${getComponentDisplayName(
```
GrandPoohBear: How married are we to `withReactNavigationAutotrack`?  For one thing, the word isn't right (should be autocapture), but for another it doesn't _read_ properly in English when you see it in a hierarchy!  You have to invert the order of the tokens in your head to make any sense of it.  What if we had this displayName be something like `HeapAutocaptured`, so that when you read the hierarchy it says `HeapAutocaptured(Main Menu)` or somesuch.

js/autotrack/heapIgnore.js
```diff
@@ -36,7 +38,9 @@ export const withHeapIgnore = (IgnoredComponent, heapIgnoreConfig) => {
   }
 
   // :TODO: (jmtaber129): Change this to 'withHeapIgnore(<IgnoredComponent name>).
```
GrandPoohBear: Looks like you can check this off your list :-)

js/autotrack/reactNavigation.js
jmtaber129: This seems to be a relatively common naming convention for HOCs (e.g. `withRouter`, `withNavigation`, etc.), so I'd prefer we just stick with this

PR #148, opened 2019-10-08
[HEAP-11501] Use iOS Framework track methods

jmtaber129:
## Description
* Upgrade the vendored iOS Heap dependency to 6.5.0-alpha.1
* Switch the native iOS methods the bridge uses from custom track to the framework track methods provided in Heap iOS v6.5.0
* Change the `type` field in navigation events to `action`, since `type` is reserved for the event name
* Update tests + test utils to use `source` and `sprops` pixel params.

This will introduce inconsistency between how iOS and Android events are sent and processed server-side, so merging into `feature/rnfcl` for now.

## Test Plan
Updated + ran detox tests, and observed correct events coming to kafka.

## Checklist
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (will be updated right before `feature/rnfcl` is merged into develop)

### DIFF
--- a/js/autotrack/reactNavigation.js
+++ b/js/autotrack/reactNavigation.js
@@ -29,7 +29,7 @@ export const withReactNavigationAutotrack = track => AppContainer => {
 
       track(EVENT_TYPE, {
         path: initialPageviewPath,
-        type: INITIAL_ROUTE_TYPE,
+        action: INITIAL_ROUTE_TYPE,
       });
     }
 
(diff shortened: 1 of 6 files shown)


js/autotrack/reactNavigation.js
```diff
@@ -29,7 +29,7 @@ export const withReactNavigationAutotrack = track => AppContainer => {
 
       track(EVENT_TYPE, {
         path: initialPageviewPath,
-        type: INITIAL_ROUTE_TYPE,
```
GrandPoohBear: Remind me - what's the motivation for this change?  Consistency with something else, or is `type` a special keyword somewhere?

jmtaber129: `type` represents the event type (param `t` in the pixel).  So if we pick a `type` field from `sprops` and throw it into the non-custom part of the event getting written to kafka, this field would overwrite the *real* `type`.  This used to not be an issue, since `type` was treated as a custom prop.

PR #149, opened 2019-10-08
[HEAP-11507] Revise names of event types + props

jmtaber129:
## Description
Revised event type/prop names:
* `touchableHandlePress`/`touchableHandleLongPress` -> `touch`
* Added `isLongPress` to `touch` events to differentiate
* `_handleChange` -> `change`
* `touchableHierarchy` -> `hierarchy`
* All other event type/prop names converted from `lowerCamelCase` to `snake_case`.

## Test Plan
Updated existing tests, and observed incoming pixel requests.

## Checklist
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (will be updated right before feature/rnfcl is merged into develop)

### DIFF
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -323,10 +323,10 @@ describe('Basic React Native and Interaction Support', () => {
     it("should autotrack 'TouchableOpacity's", async () => {
       const expectedHierarchy = `${BASICS_PAGE_TOP_HIERARCHY}TouchableOpacity;[testID=touchableOpacityText];|`;
       const expectedTargetText = 'Touchable Opacity Foo';
-      await rnTestUtil.assertAutotrackHierarchy('touchableHandlePress', {
-        touchableHierarchy: expectedHierarchy,
-        targetText: expectedTargetText,
-        screenName: 'Basics',
+      await rnTestUtil.assertAutotrackHierarchy('touch', {
+        hierarchy: expectedHierarchy,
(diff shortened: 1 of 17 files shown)


e2e/basics.spec.js
```diff
@@ -323,81 +323,81 @@ describe('Basic React Native and Interaction Support', () => {
...
 
     it("should autotrack 'TouchableHighlight's", async () => {
       const expectedHierarchy = `${BASICS_PAGE_TOP_HIERARCHY}TouchableHighlight;[testID=touchableHighlightText];|`;
       const expectedTargetText = 'Touchable Highlight';
-      await rnTestUtil.assertAutotrackHierarchy('touchableHandlePress', {
-        touchableHierarchy: expectedHierarchy,
-        targetText: expectedTargetText,
-        screenName: 'Basics',
+      await rnTestUtil.assertAutotrackHierarchy('touch', {
+        hierarchy: expectedHierarchy,
```
jnatkins: Is hierarchy, without any sort of qualifier, going to cause any potential confusion anywhere? react_hierarchy?

jmtaber129: This will be the only hierarchy coming from the `react_native` lib, so I wouldn't think so, but I'll defer to you

jnatkins: If you don't see any concern, then I don't see any concern. Thanks!

PR #150, opened 2019-10-09
[HEAP-11523] Send screen props as contextual framework props

jmtaber129:
## Description
For manual track calls, include screen props (`screen_name`, `path`) as contextual source props.

## Test Plan
Updated existing tests

## Checklist
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (will be updated right before feature/rnfcl is merged into develop)

### DIFF
--- a/e2e/basics.spec.js
+++ b/e2e/basics.spec.js
@@ -74,12 +74,11 @@ describe('Basic React Native and Interaction Support', () => {
   describe(':ios: Bridge API', () => {
     it('should call first track', async () => {
       await rnTestUtil.assertIosPixel(
-        { a: '2084764307', t: 'pressInTestEvent1' },
+        { a: '2084764307', t: 'pressInTestEvent1', sprops: [ 'path', 'Basics', 'screen_name', 'Basics' ] },
         event => {
           return !(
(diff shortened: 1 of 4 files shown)


e2e/basics.spec.js
```diff
@@ -74,51 +74,63 @@ describe('Basic React Native and Interaction Support', () => {
   describe(':ios: Bridge API', () => {
     it('should call first track', async () => {
       await rnTestUtil.assertIosPixel(
-        { a: '2084764307', t: 'pressInTestEvent1' },
+        {
+          a: '2084764307',
+          t: 'pressInTestEvent1',
+          sprops: ['path', 'Basics', 'screen_name', 'Basics'],
```
GrandPoohBear: Do these get explicitly sorted by key at some point when it's converted from a map to an array?  I'm concerned these could get flickery down the line...

jmtaber129: Yea, I think so (either when it's converted to pixel format, or within the test utils).  I wouldn't be too concerned, since this will change once the iOS message format changes happen.

jmtaber129:
Re: unit tests - Ask and you shall receive; added assertions for these to the `Heap.js` unit tests.

PR #155, opened 2019-11-11
Rename conflicting field names

jmtaber129:
## Description
Renames hierarchy to react_hierarchy and path to screen_path due to conflicts with the web `hierarchy` and `path` properties on the frontend.  This was originally expected to be a non-issue, since these fields would be from RN, and the web fields would be from web, but this doesn't seem to be the case.

## Test Plan
Updated test cases

## Checklist
- [X] Detox tests pass
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (n/a)


jmtaber129:
* I'll talk to capture
* No migration needed; this is getting merged into the unreleased feature branch.

PR #159, opened 2019-12-02
[HEAP-12315] Delegate React Navigation onNavigationStateChange

jmtaber129:
## Description
Currently, if an `onNavigationStateChange` handler is passed to the React Navigation autocapture HOC, that handler will get called instead of the handler defined within the HOC.  So if someone is manually tracking screenviews using the [instructions provided by React Navigation](https://reactnavigation.org/docs/en/screen-tracking.html), our React Navigation autocapture won't work.

Delegate to the `onNavigationStateChange` passed to the HOC within the HOC's `onNavigationStateChange` handler.

## Test Plan
Changed test app to pass a `onNavigationStateChange` handler to the HOC.

## Checklist

### DIFF
--- a/examples/TestDriver/App.js
+++ b/examples/TestDriver/App.js
@@ -17,7 +17,9 @@ export default class App extends React.Component {
       <Provider store={store}>
         <TopNavigator
           ref={NavigationService.setTopLevelNavigator}
-          onNavigationStateChange={() => {}} // Checks that manually instrumenting screen tracking works.
+          // Check that manually instrumenting screen tracking doesn't break autocapture by passing in a no-op function as the
+          // 'onNavigationStateChange' handler.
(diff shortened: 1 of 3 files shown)


examples/TestDriver/App.js
```diff
@@ -15,7 +15,10 @@ export default class App extends React.Component {
   render() {
     return (
       <Provider store={store}>
-        <TopNavigator ref={NavigationService.setTopLevelNavigator} />
+        <TopNavigator
+          ref={NavigationService.setTopLevelNavigator}
+          onNavigationStateChange={() => {}} // Checks that manually instrumenting screen tracking works.
```
GrandPoohBear: It might be good to expand on this comment, since the context will get lost pretty quickly - i.e. why does having a NOOP function there check what it checks?

PR #168, opened 2019-12-19
Strip reserved characters from hierarchies

jmtaber129:
## Description
Certain characters like `@`, `|`, and `;` are used as delimiters in hierarchies.  Now that we're supporting hierarchy matchers in event definitions, it's _especially_ important that these are only used as delimiters in the hierarchy (i.e. not present in prop or component names).  Strip these characters from props and component names.

~Needs rebase once https://github.com/heap/react-native-heap/pull/166 lands~

## Test Plan
Updated unit tests, and ran detox tests.

## Checklist


js/util/extractProps.ts
```diff
@@ -106,10 +110,11 @@ export const extractProps = (
     if (
       flattenedProps[key] !== null &&
       flattenedProps[key] !== undefined &&
-      typeof flattenedProps[key] !== 'function'
+      typeof flattenedProps[key] !== 'function' &&
+      !containsReservedCharacter(key)
     ) {
       // Remove all brackets from string.
```
GrandPoohBear: This comment can probably just go away now...

PR #174, opened 2020-01-07
Telemetry metadata enrichment

jmtaber129:
## Description
Enrich events with telemetry metadata.  Specifically, the RNHeap SDK version and whether the React Navigation HOC has been instantiated.

## Test Plan
Update e2e tests

## Checklist
- [] ~If this is a bugfix/feature, the changelog has been updated~ (N/A)

### DIFF
--- a/js/util/navigationUtil.ts
+++ b/js/util/navigationUtil.ts
@@ -25,7 +25,7 @@ export default class NavigationUtil {
     return null;
   }
 
-  static isEnabled(): boolean {
+  static isHocEnabled(): boolean {
     return !!this.heapNavRef;
   }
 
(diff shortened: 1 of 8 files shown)


js/util/navigationUtil.ts
```diff
@@ -25,6 +25,10 @@ export default class NavigationUtil {
     return null;
   }
 
+  static isEnabled(): boolean {
```
GrandPoohBear: Is "enabled" actually what we're tracking here?  I thought it was whether we were using an HOC.  Are they two ways to say the same thing?

jmtaber129: This is effectively "has the dev enabled screenview autocapture via the HOC?"

Maybe `isHocEnabled()` would be better?

jmtaber129:
refs HEAP-10613

PR #182, opened 2020-03-30
[HEAP-13861] Update index.d.ts typings with resetIdentity and getUserId methods

jmtaber129:
## Description
Adds TS typings for `resetIdentity()` and `getUserId()` methods.

## Test Plan
Ran `tsc index.d.ts` to check compilation.

## Checklist
- ~[ ] Detox tests pass (only Heap employees are able run these)~ N/A.


jmtaber129:
Yep, the only other methods not included in these typings are the should-be internal autocapture hooks used by the instrumentor.

PR #187, opened 2020-04-10
[HEAP-14089] Bump iOS dependency to version 6.7.0

jmtaber129:
## Description
Bump Heap iOS SDK dependency from version 6.5.3 to 6.7.0.

## Test Plan
Detox tests

## Checklist


jmtaber129:
No, since the JS-side API hasn't changed.  If we decide to add these native APIs to the JS API bridge, `index.d.ts` would be updated in that PR.

PR #191, opened 2020-05-29
[HEAP-15381] Improve Support for HeapIgnore in minified apps [S]

jmtaber129:
## Description
Improve support for HeapIgnore in minified apps by changing the `HeapIgnore` component to a class, explicitly set the component's `displayName`, and check against the class's displayName, rather than a string literal.

## Test Plan
Manually tested on a minified app.

## Checklist


jmtaber129:
Because of e2e testing dependencies on the internal repo, getting E2E tests (minified or otherwise) into CI is non-trivial

PR #199, opened 2020-07-08
[HEAP-16352] Alpha support for Touchables in React Native 0.62+

jmtaber129:
Touchable support for React Native 0.62+.

This involves adding instrumentation to the new RN Pressability class's constructor, copying the config, and wrapping the config's onPress and onLongPress methods.

This PR is intended for an alpha release, so testing isn't as good as it should be.  These two commits were cherry-picked from https://github.com/heap/react-native-heap/pull/198 to keep develop clean.

Still need to run detox tests for <0.62 by cherry-picking these commits onto https://github.com/heap/react-native-heap/pull/192, but otherwise ready for review.

## Test Plan
Detox tests for android pass for the 0.62 app on https://github.com/heap/react-native-heap/pull/198, but a detox issue is preventing iOS tests from running.  Because this only changes instrumentation on the JS side, test results _shouldn't_ vary between android and iOS, but this should be tested on iOS before including this in a stable release.

## Checklist
- [ ] Detox tests pass (only Heap employees are able run these) (see above)

### DIFF
--- a/instrumentor/test/index.js
+++ b/instrumentor/test/index.js
@@ -78,7 +78,7 @@ describe('autotrack instrumentor plugin', () => {
     assert.equal(actual, expected);
   });
 
-  it('non-pressability should be instrumented', () => {
+  it('non-pressability should not be instrumented', () => {
     var actual = getActualTransformedFile('is-not-pressability');
     var expected = getExpectedTransformedFile('is-not-pressability');
     assert.equal(actual, expected);
--- a/instrumentor/test/index.js
+++ b/instrumentor/test/index.js
@@ -71,6 +71,18 @@ describe('autotrack instrumentor plugin', () => {
     var expected = getExpectedTransformedFile('not-text-input');
     assert.equal(actual, expected);
   });
+
+  it('pressability should be instrumented', () => {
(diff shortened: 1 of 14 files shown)


instrumentor/test/index.js
```diff
@@ -71,6 +71,18 @@ describe('autotrack instrumentor plugin', () => {
...
     assert.equal(actual, expected);
   });
+
+  it('pressability should be instrumented', () => {
+    var actual = getActualTransformedFile('is-pressability');
+    var expected = getExpectedTransformedFile('is-pressability');
+    assert.equal(actual, expected);
+  });
+
+  it('non-pressability should be instrumented', () => {
```
kamalmarhubi, line 81:
```suggestion
  it('non-pressability should not be instrumented', () => {
```

jmtaber129:
Chatted with @kamalmarhubi synchronously about his concerns, and they mentioned feeling comfortable approving if I added some links to the RN source, but they didn't approve before signing off for the day.

In the interest of getting this landed, I'm going to merge as-is, and solicit their post-merge approval tomorrow.

PR #206, opened 2020-07-22
Include react native version with other telemetry props [S]

jmtaber129:
## Description
Include a `react_native_version` telemetry metadata prop on React Native events.  Note that this will _not_ work on React Native versions below 0.61, since that's the first version with the `Platform.constants` API (see tags associated with https://github.com/facebook/react-native/commit/7fd08e146184432731ec5d5ba210e352690dc569).

## Test Plan
* Ran Android e2e tests to confirm this doesn't lead to any regressions on older versions of RN.
* Ran this on an android app with a newer version of React Native, and confirmed that `react_native_version` is included as a source property with an accurate value.

## Checklist
- [X] Detox tests pass (only Heap employees are able run these) (ran Android; iOS shouldn't be any different)
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (not a feature)

### DIFF
--- a/js/__tests__/Heap.spec.js
+++ b/js/__tests__/Heap.spec.js
@@ -63,6 +63,7 @@ describe('The Heap object', () => {
     NavigationUtil.getScreenPropsForCurrentRoute.mockImplementation(() => {
       return {
         is_using_react_navigation_hoc: false,
+        react_native_version: null,
         screen_path: 'Basics::Foo',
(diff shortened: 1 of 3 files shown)


jmtaber129:
Our test app is currently running RN 0.57, which doesn't support this functionality.  We'll add a test for this once we use a test app with RN 0.61+

PR #207, opened 2020-07-23
Make Touchable autocapture/instrumentation more robust on RN 0.62+

jmtaber129:
## Description
This PR does a few things:
* Reverts https://github.com/heap/react-native-heap/pull/199
* Uses a new approach to instrumentation that should be more robust than previous implementations:
  * Inject code into the React Native library such that an HOC provided by Heap wraps the instrumented component
  * The HOC passes most props through to the base component, but creates new handlers for the `onPress` and `onLongPress` handlers, for which we 1) track the interaction, and 2) delegate to the `onPress` and `onLongPress` handlers passed into the rendered HOC.

So, for a `TouchableOpacity` being instrumented, we'd take code that looks like
```
class TouchableOpacity extends React.Component {...}
```
and turn it into
```
const Heap = require('@heap/react-native-heap');
const TouchableOpacity = Heap.withHeapTouchableAutocapture(class TouchableOpacity extends React.Component {...});
```

This approach should make it our current instrumentation more robust (since it uses public RN component APIs instead of internal RN code structure), and should make it easier to fix broken instrumentation in the future (this approach can probably be applied to most components)

## Test Plan
* Ran Android e2e detox tests against our test app running RN 0.57
* Manually tested against a separate test app running RN 0.62 (outside this repo)

## Checklist


js/autotrack/touchables.js
```diff
@@ -29,44 +30,45 @@ export const autotrackPress = track => (eventType, componentThis, event) => {
...
+  )})`;
 
-      pressabilityConfig.onPress(event);
-    };
-  }
-
-  if (newConfig.onLongPress) {
-    newConfig.onLongPress = event => {
-      pressHandler(event, track, true);
+  const forwardRefHoc = React.forwardRef((props, ref) => {
```
jamesnghiem, line 67: forwardRef essentially just passes the ref to the children right?

jmtaber129: Yep, it basically makes it such that you can forward the ref to some internal component (in this case, so you get the ref for the touchable component, and not the HOC wrapping the touchable component)

PR #208, opened 2020-07-24
HOC instrumentation for TextInput on RN 0.62+

jmtaber129:
## Description
This is similar to https://github.com/heap/react-native-heap/pull/207 in that it adds HOC-based instrumentation for `TextInput` components.

Starting with RN 0.62, the internal implementation of the React Native `TextInput` component uses a functional component and React Hooks (this used to be a stateful class component).  Because of this, support for this component in RN 0.62+ would require new instrumentation (not as complicated as non-HOC `Touchable` support, but still non-trivial), making this instrumentation fix a candidate for the HOC-based instrumentation approach.

The instrumentation looks for a function named `InternalTextInput` and wraps it with an HOC.  I.e.
```
function InternalTextInput(props: Props): React.Node {}
```
becomes
```
const Heap = require('@heap/react-native-heap').default;

const InternalTextInput = Heap.withHeapTextInputAutocapture(function InternalTextInput(props) {});
```
(see the test cases).

This is based on https://github.com/heap/react-native-heap/pull/207, so when https://github.com/heap/react-native-heap/pull/207 lands, this branch will need to be rebased onto `develop`, and this PR's base will need to be changed to `develop`.

## Test Plan
* Added instrumentation test case
* Ran e2e tests on the 0.57 test app to ensure no regressions
* Manually triggered text change on 0.62 test app, and verified debounced text change events.

## Checklist
- [X] Detox tests pass (only Heap employees are able run these) (android only)

### DIFF
--- a/instrumentor/src/index.js
+++ b/instrumentor/src/index.js
@@ -351,16 +351,16 @@ const instrumentTextInputHoc = path => {
     path.node.body
   );
 
+  const hocIdentifier = t.identifier('withHeapTextInputAutocapture');
+
   const autotrackExpression = t.callExpression(
-    t.memberExpression(
-      t.identifier('Heap'),
-      t.identifier('withHeapTextInputAutocapture')
-    ),
+    t.memberExpression(t.identifier('Heap'), hocIdentifier),
     [equivalentExpression]
   );
 
   const replacement = buildInstrumentationHoc({
     COMPONENT_ID: path.node.id,
+    HOC_IDENTIFIER: hocIdentifier,
     HOC_CALL_EXPRESSION: autotrackExpression,
   });
 
--- a/instrumentor/src/index.js
+++ b/instrumentor/src/index.js
@@ -340,6 +340,33 @@ const instrumentTouchableHoc = path => {
   path.replaceWithMultiple(replacement);
 };
 
+const instrumentTextInputHoc = path => {
+  if (path.node.id.name !== 'InternalTextInput') {
+    return;
+  }
+
+  const equivalentExpression = t.functionExpression(
+    path.node.id,
+    path.node.params,
+    path.node.body
+  );
+
+  const autotrackExpression = t.callExpression(
+    t.memberExpression(
+      t.identifier('Heap'),
+      t.identifier('withHeapTextInputAutocapture')
+    ),
+    [equivalentExpression]
+  );
+
+  const replacement = buildInstrumentationHoc({
+    COMPONENT_ID: path.node.id,
+    HOC_CALL_EXPRESSION: autotrackExpression,
+  });
+
+  path.replaceWithMultiple(replacement);
+};
+
 function transform(babel) {
   return {
     visitor: {
@@ -354,6 +381,9 @@ function transform(babel) {
       },
       ClassDeclaration(path) {
         instrumentTouchableHoc(path);
+      },
+      FunctionDeclaration(path) {
+        instrumentTextInputHoc(path);
       },
     },
   };
--- a/js/autotrack/textInput.js
+++ b/js/autotrack/textInput.js
@@ -1,5 +1,10 @@
-import { getBaseComponentProps } from './common';
+import * as React from 'react';
+import hoistNonReactStatic from 'hoist-non-react-statics';
 import * as _ from 'lodash';
+
+import { getBaseComponentProps } from './common';
+import { bailOnError } from '../util/bailer';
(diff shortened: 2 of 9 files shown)


js/autotrack/textInput.js
```diff
@@ -37,3 +42,57 @@ const debouncedAutocaptureTextInputChange = track => (
...
+    );
+
+    render() {
+      const { forwardedRef, onChange, ...rest } = this.props;
+
+      return (
+        <TextInputComponent
+          ref={forwardedRef}
+          onChange={e => {
+            this.autocaptureTextInputChangeWithDebounce('text_edit', this, e);
```
jamesnghiem, line 71: would it be worthwhile to move `text_edit` to a constants file (assuming that its used in various other places)

js/autotrack/textInput.js
```diff
@@ -37,3 +42,57 @@ const debouncedAutocaptureTextInputChange = track => (
...
+// :TODO: (jmtaber129): Consider other workarounds to this, like setting the display name of instrumented components when they're exported
+// from the React Native lib (via instrumentation).
+const NoopTextInput = props => {
+  return props.children;
+};
+
+NoopTextInput.displayName = 'TextInput';
+
+export const withHeapTextInputAutocapture = track => TextInputComponent => {
+  class HeapTextInputAutocapture extends React.Component {
```
jamesnghiem, line 59: wondering if this needs to be a class component since its not really stateful

instrumentor/src/index.js
```diff
@@ -295,6 +300,43 @@ const instrumentStartup = path => {
...
+  'TouchableWithoutFeedback',
+  'TouchableHighlight',
+];
+
+const instrumentTouchableHoc = path => {
+  if (!TOUCHABLE_COMPONENTS.includes(path.node.id.name)) {
+    return;
+  }
+
+  // 'path.node' represents a class *declaration*, so we need to convert 'path.node' to a class *expression* before we can pass it as an
```
timothyandrew, line 315: What is the difference between `declaration` and `expression` here? (An example would be nice! 🙂)

js/autotrack/touchables.js
```diff
@@ -24,3 +29,46 @@ export const autotrackPress = track => (eventType, componentThis, event) => {
...
+          }}
+          {...rest}
+        >
+          {this.props.children}
+        </TouchableComponent>
+      );
+    }
+  }
+
+  HeapTouchableAutocapture.displayName = `withHeapTouchableAutocapture(${getComponentDisplayName(
```
timothyandrew, line 63: nit: capitalization

```suggestion
  HeapTouchableAutocapture.displayName = `WithHeapTouchableAutocapture(${getComponentDisplayName(
```

js/autotrack/textInput.js
```diff
@@ -37,3 +42,57 @@ const debouncedAutocaptureTextInputChange = track => (
...
+          }}
+          {...rest}
+        >
+          {this.props.children}
+        </TextInputComponent>
+      );
+    }
+  }
+
+  HeapTextInputAutocapture.displayName = `withHeapTextInputAutocapture(${getComponentDisplayName(
```
timothyandrew, line 83: nit: capitalization

```suggestion
  HeapTextInputAutocapture.displayName = `WithHeapTextInputAutocapture(${getComponentDisplayName(
```

js/autotrack/textInput.js
jmtaber129: These event types are kinda all over the place in the repo (some are in the instrumentor, others are in the library), so I think I'd prefer to leave this sort of thing for another time.

js/autotrack/textInput.js
jmtaber129: The `autocaptureTextInputChangeWithDebounce` is effectively state (not the traditional React `state`, but the sort of object state you'd want a class component for).

instrumentor/src/index.js
jmtaber129: It's kinda difficult to explain without looking at the AST the code parses into, since a `ClassDeclaration` and `ClassExpression` _look exactly the same in code_.

I _think_ that a declaration declares the name of the class as an in-scope symbol, whereas a class expression is just a class with a name (the class doesn't actually get declared as an in-scope thing).

emodventures:
A quick note that this PR breaks use of ref on TextInput components if the react-native-heap instrumentor is enabled in the babel configuration. (reproduced with React Native 0.63.2)

PR #210, opened 2020-07-28
[HEAP-16942] Bump version to 0.13.0-alpha2

jmtaber129:
## Description
Bump version to 0.13.0-alpha2 with a fix related to RN 0.62 support changes.

Since this is only an alpha, and the TestDriver iOS test app isn't currently working on my machine (Catalina), I only ran Android detox tests.

## Checklist
- [ ] Detox tests pass (only Heap employees are able run these) (Android only, see above)
- [ ] If this is a bugfix/feature, the changelog has been updated (Alpha release, so no changelog update)

### DIFF
--- a/examples/TestDriver/package.json
+++ b/examples/TestDriver/package.json
@@ -8,7 +8,7 @@
     "preinstall": "./preinstall_pack_lib.sh"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


jmtaber129:
We update the changelog for stable releases, but not for pre-releases like this alpha

PR #213, opened 2020-08-03
[HEAP-16909] React Navigation 5 Support

jmtaber129:
## Description
Due to API changes between v4 and v5 of the React Navigation library, our `withReactNavigationAutotrack` HOC would not work for apps running React Navigation 5.  This PR adds support for React Navigation 5.

## Test Plan
Ran E2E detox tests on Android to test for regressions on React Navigation 3/4.
TODO: Manual testing of React Navigation 5 on the new test app.

## Checklist

### DIFF
--- a/js/autotrack/reactNavigation.js
+++ b/js/autotrack/reactNavigation.js
@@ -12,7 +12,7 @@ const EVENT_TYPE = 'react_navigation_screenview';
 const INITIAL_ROUTE_TYPE = 'Heap_Navigation/INITIAL';
 
 export const withReactNavigationAutotrack = track => AppContainer => {
-  const captureNavigationStateChange = bailOnError((prev, next, action) => {
+  const captureOldNavigationStateChange = bailOnError((prev, next, action) => {
     const { screen_path: prevScreenRoute } = NavigationUtil.getActiveRouteProps(
       prev
     );
@@ -30,6 +30,7 @@ export const withReactNavigationAutotrack = track => AppContainer => {
 
   class HeapNavigationWrapper extends React.Component {
     topLevelNavigator = null;
+    currentPath = null;
 
     setRef(ref, value) {
       if (typeof ref === 'function') {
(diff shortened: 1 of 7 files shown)


js/autotrack/reactNavigation.js
```diff
@@ -78,14 +110,33 @@ export const withReactNavigationAutotrack = track => AppContainer => {
                 'Heap: React Navigation is instrumented for autocapture.'
               );
               this.topLevelNavigator = navigatorRef;
```
jamesnghiem, line 112: i'm wondering why use class variables for this instead of the class's `state`? I'm not sure if there's differences between RN and React in terms of class state and render lifecycles so genuinely curious

js/util/navigationUtil.ts
```diff
@@ -10,13 +10,16 @@ export default class NavigationUtil {
     screen_path: string;
     screen_name: string;
   } | null {
-    if (
-      !(this.heapNavRef && this.heapNavRef.state && this.heapNavRef.state.nav)
-    ) {
+    let rootState: any = null;
+    if (this.heapNavRef && this.heapNavRef.state && this.heapNavRef.state.nav) {
```
jamesnghiem, line 14: What is the `heapNavRef` exactly? (when is it set, what component/element are you creating a ref on, why do you need the ref). Sorry bout all the q's 😅

js/autotrack/reactNavigation.js
```diff
@@ -78,14 +110,33 @@ export const withReactNavigationAutotrack = track => AppContainer => {
...
+                this.trackInitialRouteForState(
+                  this.topLevelNavigator.state.nav
+                );
+              }
             }
           })}
+          onReady={(...args) => {
+            this.captureOnReady();
+
+            if (typeof onReady === 'function') {
```
jamesnghiem, line 124: are onReady and onStateChange built-in RN functions? Not familiar with those methods

js/autotrack/reactNavigation.js
jmtaber129: AFAIK we don't want to update `state` (thus triggering a re-render) every time the ref changes

js/util/navigationUtil.ts
jmtaber129: This is the ref to the NavigationContainer or AppContainer (depending on React Nav version).  We need it to obtain the current state of navigation (route path and name) for screenview props on events, and screenview props on screenviews

js/autotrack/reactNavigation.js
jmtaber129: These are input props for the `NavigationContainer` that the HOC is wrapping (see [`onReady`](https://reactnavigation.org/docs/navigation-container#onready) and [`onStateChange`](https://reactnavigation.org/docs/navigation-container#onstatechange))

PR #216, opened 2020-08-31
[HEAP-17696] Fix instrumentation build failure on id-less function declarations

jmtaber129:
## Description
Fixes a bug introduced with the TextInput HOC-based instrumentation in which the presence of a `FunctionDeclaration` without an `id` (as is the case for declarations like `export default function() {}`) would cause the instrumentor to crash.

Check whether `id` on `FunctionDeclaration`s is present, and add tests for this case.

## Test Plan
- Added instrumentor test case
- Ran Android detox tests

## Checklist
- [ ] ~If this is a bugfix/feature, the changelog has been updated~ (N/A)

### DIFF
--- a/instrumentor/test/fixtures/is-function-declaration-no-id/code.js
+++ b/instrumentor/test/fixtures/is-function-declaration-no-id/code.js
@@ -1,0 +1,1 @@
+export default function() {}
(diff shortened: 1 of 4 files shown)


instrumentor/test/fixtures/is-function-declaration-no-id/code.js
```diff
@@ -0,0 +1 @@
+export default function() {}
```
jmtaber129, line 1: This is a function declaration, but the function has no name/id

PR #222, opened 2020-09-17
[HEAP-17331] Disable native iOS capture by default

jmtaber129:
## Description
Add an `enableNativeTouchEventCapture` option for `heap.config.json` that defaults to `false`.  This flag determines whether events captured by the native iOS SDK (which are generally superfluous to events captured by the RN SDK, and are almost always unusable) will be processed.  This is intended to replace the need for `heapIgnore` as described in the [React Native iOS install docs](https://developers.heap.io/docs/react-native#ios-1) while ensuring that views created outside the root view (as is the case for certain modals) will also be ignored by the native iOS SDK.

In cases where parts of an app are implemented as native iOS views, `enableNativeTouchEventCapture` should be set to true, and `heapIgnore` should still be set on the root view.

This feature builds on the functionality implemented for auto-initialization (which assumed only `heapAutoInit` and `heapAppId` config props), so there may be on opportunity for a better refactor that allows for better extensibility to additional options.  I'm open to suggestions.

## Test Plan
Manually tested auto-initialization with and without `enableNativeTouchEventCapture` set, and confirmed that native iOS events did or didn't come through.

## Checklist
- [ ] ~Detox tests pass (only Heap employees are able run these)~ (N/A)

### DIFF
--- a/ios/HeapSettings.bundle/generate_settings_common.rb
+++ b/ios/HeapSettings.bundle/generate_settings_common.rb
@@ -32,14 +32,14 @@ def perform_empty_substitution(dir)
   IO.write output_path, settings
 end
 
-def get_app_id_for_config(specified_config, default_config)
-  is_auto_init = specified_config['heapAutoInit']
+def get_app_id_for_config(config_for_selected_environment, default_config)
+  is_auto_init = config_for_selected_environment['heapAutoInit']
   is_auto_init = default_config['heapAutoInit'] if is_auto_init.nil?
   is_auto_init = true if is_auto_init.nil?
 
-  return specified_config['heapAppId'] || default_config['heapAppId'] if is_auto_init
+  return config_for_selected_environment['heapAppId'] || default_config['heapAppId'] if is_auto_init
 end
 
-def get_enable_autocapture_for_config(specified_config, default_config)
-  return specified_config['enableNativeTouchEventCapture'] || default_config['enableNativeTouchEventCapture']
+def get_enable_autocapture_for_config(config_for_selected_environment, default_config)
+  return config_for_selected_environment['enableNativeTouchEventCapture'] || default_config['enableNativeTouchEventCapture']
 end
--- a/ios/HeapSettings.bundle/generate_settings_common.rb
+++ b/ios/HeapSettings.bundle/generate_settings_common.rb
@@ -2,23 +2,34 @@
 
 require 'fileutils'
 
-def perform_substitution(dir, app_id_dev, app_id_prod)
-  template = File.read File.join(dir, 'HeapSettings.plist.template')
-  output_path = File.join dir, 'HeapSettings.plist'
-
-  settings = template
-    .gsub('__HEAP_APP_ID_DEV__', app_id_dev || '')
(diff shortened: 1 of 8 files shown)


ios/HeapSettings.bundle/generate_settings_common.rb
```diff
@@ -2,23 +2,34 @@
...
+  prod = json.fetch('prod', {})
+  default = json.fetch('default', {})
+
   template = File.read File.join(dir, 'HeapSettings.plist.template')
   output_path = File.join dir, 'HeapSettings.plist'
 
   settings = template
-    .gsub('__HEAP_APP_ID_DEV__', app_id_dev || '')
-    .gsub('__HEAP_APP_ID_PROD__', app_id_prod || '')
+    .gsub('__HEAP_APP_ID_DEV__', get_app_id_for_config(dev, default) || '')
```
jamesnghiem, line 14: would it make sense to return the default empty string value as part of `get_app_id_for_config`?

jamesnghiem (approved):
one nit, but otherwise LGTM. I'm not super familiar with iOS dev though so might be worthwhile to wait on Tim's review if this is a high risk change

ios/HeapSettings.bundle/generate_settings_common.rb
```diff
@@ -28,3 +39,7 @@ def get_app_id_for_config(specified_config, default_config)
 
   return specified_config['heapAppId'] || default_config['heapAppId'] if is_auto_init
 end
+
+def get_enable_autocapture_for_config(specified_config, default_config)
```
timothyandrew, line 43: nit: `specified` is a bit vague on its own here; consider something like `config_for_selected_environment` instead.

timothyandrew (approved):
A couple notes, but LGTM! It think it's worth adding the `enableNativeTouchEventCapture` key to `examples/TestDriver/heap.config.json`, too.

These ruby scripts still seem very hacky, though; how painful are they in practice? Is it worth considering a more robust solution for RN config generation?

ios/HeapSettings.bundle/generate_settings_common.rb
jmtaber129: Hmm that would be reasonable, though I think I slightly prefer the current pattern of "`nil` if no app ID, then set to an empty string in the actual plist"

jmtaber129:
Since this config prop is false by default (and usually won't be explicitly set by devs), leaving this flag out of the `heap.config.json` in the test apps is probably preferred.  I.e. excluding this prop from the config will lead native iOS events being suppressed.

The ruby scripts are definitely painful, both from an "I don't know how to write good ruby" perspective, as well as a "testing this is difficult" perspective.  At the end of the day, they don't take _that_ much time to work with, but the main reason we scored this task relatively high was because of the current auto-init architecture for iOS.  I haven't done a ton of thinking about what else we can do (e.g. use the raw JSON in the objc auto-init logic, making the ruby logic more of a "pass-through" of the JSON, etc.), but we should probably revisit how we do this, especially if we anticipate additional RN heap configuration props.

PR #268, opened 2021-11-03
0.17.0 Release

bnickel:
**In progress**

## Description

There are four goals for this PR:

1. Fix a performance issue issue on iOS where native view hierarchies are scanned but not used.
2. Start using heap-ios-sdk 8.0.0 from Cocoapods, unpinning from a specific version.
2. Make detox tests publicly runnable using a mock web server.
3. Prepare for an actual 0.17 release.

I also touched up documentation and the typescript definitions.

Given the shift in ReactNative towards CocoaPods, I'm dropping support for anything below 0.63 as well as manual iOS installation.

## Test Plan

Test plan is to get a decent collection of tests from the old e2e tests running publicly and test with detox.

## Checklist
- [x] Detox tests pass


index.d.ts
```diff
@@ -124,7 +124,7 @@ export function getUserId(): Promise<string>;
  * @param IgnoredComponent the component to wrap with HeapIgnore.
  * @param heapIgnoreConfig the HeapIgnore configuration
  */
-export function withHeapIgnore(IgnoredComponent: React.Component, heapIgnoreConfig?: object): React.Component;
+ export function withHeapIgnore<P>(IgnoredComponent: React.JSXElementConstructor<P>, heapIgnoreConfig?: object): React.JSXElementConstructor<P>;
```
bnickel, line 127: Things like `Button` inherit `React.Component` but what we pass in is something to create the component not a component itself.

integration-tests/drivers/TestDriver063/.gitignore
```diff
@@ -34,6 +34,7 @@ local.properties
 node_modules/
 npm-debug.log
 yarn-error.log
+package-lock.json
```
apyousef, line 37: Should this actually be in gitignore? I always thought package-lock.json was supposed to be in version control, at least based on things I've read in SO.

integration-tests/drivers/TestDriver063/android/app/build.gradle
```diff
@@ -134,6 +134,9 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName "1.0"
+        
+        testBuildType System.getProperty('testBuildType', 'debug')  // This will later be used to control the test apk build type
```
apyousef, line 139: Nice

integration-tests/tests/server.ts
```diff
@@ -86,7 +80,7 @@ class CaptureServer extends EventEmitter {
   constructor() {
     super();
     this.app = express();
-    this.app.get('/h', this.onGetRequest.bind(this));
+    this.app.get('*', this.onGetRequest.bind(this));
```
apyousef: Are there other get endpoints besides `/h`?

apyousef: It looks like maybe `/api/add_user_properties` and `/api/identify_v3` could also be get requests. Is that correct?

index.d.ts
apyousef: Huh, was this a mistake that was sitting there?

docs/android-setup.md
```diff
@@ -1,45 +1,31 @@
...
+  }
+}
+```
 
-  ```xml
-  <string name="com.heapanalytics.android.buildConfigPkgName">com.your_package_name</string>
-  ```
+Add the following to `android/app/build.gradle`:
 
-and replace `com.your_package_name` with the package name from the manifest tag in `AndroidManifest.xml`.
```
apyousef, line 42: I believe this part needs to stay. The way we currently do RN does not involve use of the Gradle plugin for Android, since we're not interested in instrumenting Android UI widgets. Without the application of the plugin, there needs to be a way for the user to provide the build config package name.

docs/android-setup.md
```diff
@@ -1,45 +1,31 @@
...
+    ...
+    heapVersion = '1.9.+'
   }
-  ```
-
-If you are seeing runtime warnings at startup mentioning `Heap: Could not find BuildConfig`, add the
-following line to the resources section of `res/values/strings.xml`:
+  dependencies {
+    ...
+    classpath "com.heapanalytics.android:heap-android-gradle:${heapVersion}"
```
apyousef, line 13: So, as I understand things, RN does not use the gradle plugin. Basically we're using the business logic of the Heap Android library, and eschewing the tools we use for Android autocapture.

docs/android-setup.md
```diff
@@ -1,45 +1,31 @@
...
-you should revert any changes made by `react-native link` and try the following steps, instead.
+Run `react-native link` to link the library then add the following to `android/build.gradle`:
 
-Add the following to your `settings.gradle`:
-
-  ```groovy
-  include ':react-native-heap'
-  project(':react-native-heap').projectDir =
-      new File(rootProject.projectDir, '../node_modules/@heap/react-native-heap/android')
-  ```
```
apyousef, line 13: Is this omission correct?

docs/android-setup.md
```diff
@@ -1,45 +1,31 @@
...
-
-Then add the following to `app/build.gradle`:
-
-  ```groovy
-  implementation project(':react-native-heap')
-  ```
```
apyousef, line 19: See above question.

integration-tests/drivers/TestDriver063/android/app/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: "com.android.application"
+apply plugin: 'com.heapanalytics.android'
```
apyousef: So per the comments above, I think RN has avoided using the Android Gradle plugin.

apyousef (commented):
@bnickel I left some comments.

examples/TestDriver/ios/TestDriver/AppDelegate.m
```diff
@@ -10,24 +10,10 @@
...
-#import "Heap.h"
+#import <Heap/Heap.h>
 
 @implementation AppDelegate
 
-+ (void)load {
-  // Send events to local collector.
-  SEL setRootUrlSelector = @selector(setRootUrl:);
-  if ([[Heap class] respondsToSelector:setRootUrlSelector]) {
-    [[Heap class] performSelector:setRootUrlSelector withObject:@"http://localhost:3000"];
```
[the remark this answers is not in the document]
bnickel: It is (see https://github.com/heap/react-native-heap/pull/268/files#diff-e62c463075eecb57d4bdaddfb4ccabf5819f0bb6445976c6859df6a0d9173d78).

heap-ios-sdk 8.0.0 added an option `captureBaseUrl` which we're using instead.  There's no parity with Android (that being done with gradle options) so I'm not including it in the documentation.

integration-tests/drivers/TestDriver063/.gitignore
bnickel: With the test project in particular, I've been struggling with `package-lock.json`.

Right now, my local `package-lock.json` has the following definition:

```
    "@heap/react-native-heap": {
      "version": "file:../../heap-react-native-heap-0.17.0.tgz",
      "integrity": "sha512-KhBins/aSATHvr3SJosfh0VbQ4wsBeWko4dRStQx9RfjQCCcxzBgnjX3hLNrLNtYZzOJ8aYYZfuLlk8M8U66/g==",
      "requires": {
        "@babel/core": "^7.16.0",
        "@types/react-reconciler": "^0.26.4",
        "babel-plugin-add-react-displayname": "0.0.5",
        "babel-template": "^6.26.0",
        "babel-types": "^6.26.0",
        "flat": "^5.0.2",
        "hoist-non-react-statics": "^3.3.2"
      }
    },
```

If I make a change to one of our source files, repack the project, and wipe out `node_modules`, running `npm install` will fail to install my updated local source and instead install a cached version of `react-native-heap` that has that sha512 signature.

There may be a better way to force the install despite the changed signature, but deleting `package-lock.json` was my solution to get the tests to consistently consume the latest source.

integration-tests/drivers/TestDriver063/android/app/build.gradle
bnickel: I don't understand any of it, but it came from https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md 😄

integration-tests/tests/server.ts
bnickel: That's right.  Since all GET requests can be represented as a `URL` object, I just stash them in the same place and have a single method that gets each type: `expectPixelRequest`.  A good example of how this is used is `expectIdentify(identity: string)`:

https://github.com/heap/react-native-heap/blob/26370e65e6cdc389271b2e05ceb96fc1be733a0a/integration-tests/tests/util/server.ts#L347-L364

index.d.ts
bnickel: Yeah.  Typescript is fun. We get some same weirdness in Objective-C when you deal with passing classes around as protocol implementations.  I noticed the errors when bringing HeapIgnore.tsx from the old e2e project.

The definition is:

```
    type JSXElementConstructor<P> =
        | ((props: P) => ReactElement<any, any> | null)
        | (new (props: P) => Component<any, any>);
```

Beyond silencing the errors, this actually improves things in two ways:

1. It also removes errors on element functions like:

    ```tsx
    const Foo = ({children}) => {
      return <>{children}</>;
    };

    const IggyFoo = Heap.withHeapIgnore(Foo);
    ```

2. It makes the returned component accept the same params (`<P>`) as the original type, so you get full type checking on the element.

docs/android-setup.md
bnickel: Same.

docs/android-setup.md
bnickel: Ah, yeah, this was required for versions below 0.60.

docs/android-setup.md
bnickel: Oops, I was looking at the wrong documentation and can rip this out. The tests were passing without it so I'll revert it from the test project as well.

docs/android-setup.md
bnickel: Oops, same deal.  Will revert.

integration-tests/drivers/TestDriver063/android/app/build.gradle
```diff
@@ -194,7 +193,6 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
 
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
-    implementation "com.heapanalytics.android:heap-android-client:$heapVersion"
```
apyousef: Careful. I think we still want `heap-android-client` (the actual library, as opposed to the plugin).

apyousef: Ah sorry, I think I missed what you did in `./android/build.gradle`. In that case, I think you might be okay here actually.

integration-tests/drivers/TestDriver063/.gitignore
bnickel: I found an alternative solution with:

```sh
npm uninstall @heap/react-native-heap || true
npm install ../../heap-react-native-heap-*.tgz

apyousef (approved):

Issue #242, opened 2021-04-23
Heap's minSDKVersion of 16 isn't supported after upgrading to React Native 64


RustyDev:
```
uses-sdk:minSdkVersion 16 cannot be smaller than version 21 declared in library [com.facebook.react:react-native:0.64.0] 
```

apyousef:
@RustyDev Thank you for your patience.

If you set `android.defaultConfig.minSdkVersion` to 21 in your `app/build.gradle` file, does that resolve the issue?

PR #271, opened 2021-11-12
Add tests for 0.66, fix Android test .gitignore

bnickel:
## Description

There are no code changes in this PR, just more tooling stuff:

- I had a bad line in .gitignore for the RN 0.63 test driver so Android source was not included.
- I added a test driver for RN 0.66 and added options in test.sh to pick the driver and limit test actions.
- The 0.66 test demonstrated issue #242 so I implemented what I believe is a fix, but I don't really get why Android would fail on this.  I'll defer to @apyousef for it it's the right fix.

## Test Plan

Automated tests pass.

## Checklist
- [x] Detox tests pass


android/build.gradle
```diff
@@ -55,7 +55,7 @@ android {
         targetCompatibility 1.8
     }
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion safeExtGet('minSdkVersion', 16)
```
bnickel, line 58: @apyousef This is the only interesting thing here.  I believe what's happening here is that RN 0.64+ has a higher minSdkVersion than 16, so we fail when linking against it.

integration-tests/drivers/TestDriver063/.gitignore
```diff
@@ -59,5 +59,5 @@ buck-out/
 /ios/Pods/
 
 # From test script
-src/
-tests/
+/src/
```
bnickel, line 62: This was wiping out /android/app/src/ 😦 What'e even the point of `*/foo` in .gitignore if `foo` does the same thing?

apyousef (approved):
@bnickel I'm so sorry this took so long to approve. The Android changes look fine to me, as does the rest of the PR.

PR #285, opened 2022-03-21
iOS Event Visualizer support

bnickel:
## Description

This PR adds support for Event Visualizer on iOS.  Most of the work was done in the Heap 8.2 release, but this covers some of the finer details:

- Pairing can be turned on using the React Native Dev Menu.
- When paired using the menu, pageview events are delayed by 500ms.  Both ReactNavigation transitions and RNSScreen view appearances fire at the beginning of a transition so letting the view settle gives us the best chance for a good screenshot.

## Test Plan

This was tested manually on an Expo app and a ReactNative 0.68 app.

## Checklist
- [x] Detox tests pass


ios/RNHeap.m
```diff
@@ -14,7 +22,30 @@ @implementation RNHeap
...
+    return dispatch_get_main_queue();
+}
+
+static void(^delayedPageview)(void);
+
+/// This function delays a pageview until we think the animation has completed.
+/// Since the state change fires at the start of the animation, we wait half a second for the transiton to complete.
++ (void)scheduleDelayedPageview:(void(^)(void))block
+{
+    [self consumeDelayedPageview];
```
apyousef, line 34: I'm not sure I completely understand this line. Is this to flush out a prior pageview?

apyousef (commented):
@bnickel I had one question.

ios/RNHeap.m
bnickel: I went ahead and simplified things so there's no need for this.  Previously, I was thinking I could consume pageviews early when `viewDidAppear:` was called, but it just ends up firing early so there was no use for the complexity.

PR #303, opened 2022-04-27
Android Version Updates

MCeley:
## Description
This PR updates the Heap Android library to version 1.10.+ ~~, Gradle to 7.2, and the Android Gradle Plugin to 7.1.3.~~

## Test Plan
All integration tests passed using TestDriver066.
Tested manually in ExpoEjectedApp from the React Native samples repo.

## Checklist
- [x] Detox tests pass


android/build.gradle
```diff
@@ -42,7 +42,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:7.1.3'
```
bnickel, line 45: Is there a reason for bumping the gradle version?  If I run a recently created Expo project, I get warnings like:

```
> Configure project :expo-file-system
WARNING: Configuration 'testApi' is obsolete and has been replaced with 'testImplementation'.
It will be removed in version 5.0 of the Android Gradle plugin.
For more information, see http://d.android.com/r/tools/update-dependency-configurations.html.
```

```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.9/userguide/command_line_interface.html#sec:command_line_warnings
```

This project doesn't use any of Heap's gradle instrumentation, so I'd be inclined to keep this as low as possible to avoid conflicts with other projects.

MCeley: It's not necessary to bump. This was just a proactive bump in order to keep up with Android tools versions, but I can drop it back down for the sake of compatibility with React Native tooling.

bnickel (approved):
lgtm, the locks changed so much because you're on a newer version of npm than I was.  It looks like that's being used elsewhere, so :+1:

PR #327, opened 2022-08-10
[HEAP-33666] Changed exception to warning

semmerich-heap:
## Heap Customer Support

Thanks for using Heap's React Native SDK! If you're having any issues, please reach out to customer support at <support@heap.io>. For the best service, include "React Native" in the subject and your app id or customer email address in the body. Also, feel free to file a github issue or open a PR! If you do so, be sure to include a link in your request to customer support. Our engineering team checks for new PRs and github issues and tries to respond as soon as possible.

## Description
What does this PR add/fix/change?
 - Modified the android bridge to log a warning instead of throw an error. 
Is there anything reviewers should pay special attention to?
Are there any breaking changes?

## Test Plan
How were these changes tested?
Were additional test cases added?
Was there manual testing?

## Checklist
- [X] Detox tests pass
- [ ] If this is a bugfix/feature, the changelog has been updated

### DIFF
--- a/android/src/main/java/com/heapanalytics/reactnative/RNHeapLibraryModule.java
+++ b/android/src/main/java/com/heapanalytics/reactnative/RNHeapLibraryModule.java
@@ -12,6 +12,8 @@ import com.heapanalytics.android.internal.HeapImpl;
 import java.util.HashMap;
 import java.util.Map;
 
+import android.util.Log;
+
 public class RNHeapLibraryModule extends ReactContextBaseJavaModule {
 
   private final ReactApplicationContext reactContext;
@@ -81,7 +83,7 @@ public class RNHeapLibraryModule extends ReactContextBaseJavaModule {
         // The JS bridge will flatten maps and arrays in a uniform manner across both
         // platforms.
         // If we get them at this point, we shouldn't continue.
-        throw new RNHeapException("Property objects must be flattened before being sent across the JS bridge.");
+        Log.w("Property objects must be flattened before being sent across the JS bridge. If you get this warning please inspect for non-flattenable objects being sent to Heap");
       }
     }
     return stringMap;


android/src/main/java/com/heapanalytics/reactnative/RNHeapLibraryModule.java
```diff
@@ -81,7 +83,7 @@ private static Map<String, String> convertToStringMap(ReadableMap readableMap) {
         // The JS bridge will flatten maps and arrays in a uniform manner across both
         // platforms.
         // If we get them at this point, we shouldn't continue.
-        throw new RNHeapException("Property objects must be flattened before being sent across the JS bridge.");
+        Log.w("Property objects must be flattened before being sent across the JS bridge. If you get this warning please inspect for non-flattenable objects being sent to Heap");
```
bnickel, line 86: Do we have a payload we know causes an error today?  Would it be possible to use that to add an integration test? 

I'd take a look at the integration tests for `getUserId` and `getSessionId`.

The event source could be:

```
      <Button
        title="Send Custom Event"
        onPress={async () =>
          Heap.track('custom-event', { a: {b: null, c: [1,2,3], d: new (class Test {}) } })
        }
      />
```

This produces the event properties `{ 'a.b': null, 'a.c.0': 1, 'a.c.1': 2, 'a.c.2': 3, 'a.d': Test {} }`, which I believe _could_ produce a crash today.  **We should verify if we don't know.**

Integration tests would then validate with:

```
  describe('track', () => {
     it('should serialize properties without crashing', async () => {
       await tapButton('Send Custom Event');

       const message = await tools.server.expectSourceCustomEventWithProperties(
         'custom-event',
         'a.c.0': '1',
         'a.c.1': '2',
         'a.c.1': '3',
       );
     });
   });
```

bnickel: Also, we may want to include the name of the key so developers know where to look.

bnickel (approved):
This looks good.  A couple notes but they're not blockers if this needs to go out fast:

1. I would add a test to verify that (A) we know what the problem was and (B) that we fixed it. (Comments inline.)
2. The changelog should also be updated but that could land in the PR.

Issue #374, opened 2023-03-08
Pod install failing: `Unable to validate due to exception: undefined method`


rcorrie:
SDK version: `0.22.1`

Cocoapods version: `1.12.0`

```
[!] The `react-native-heap` pod failed to validate due to 1 error:

    - ERROR | source: Unable to validate due to exception: undefined method `=~' for {:type=>"git", :url=>"git+https://github.com/heap/react-native-heap"}:Hash
```

huynhtastic:
I believe this is an issue with cocoapods `1.12.0`. I didn't get this error until I ran `brew upgrade cocoapods`. Would love a fix for this.

amq:
Ruby 3.0.5 works, 3.2.1 doesn't

PR #375, opened 2023-03-23
Resolve install issue with ruby >= 3.1

amq:
## Description

Fixes https://github.com/heap/react-native-heap/issues/374

## Test Plan

Tested locally by running `bundle exec pod install` with ruby 3.2.1 and ruby 3.0.5

## Checklist

- [x] Detox tests pass


MCeley:
Fixed in #377

PR #401, opened 2023-08-19
[HEAP-46713] Fix switch capture

bnickel:
## Description

This fixes the capture of `<Switch>` components on React Native 0.66.

In 0.66, the definition changed from a class definition to a variable initialized with `React.forwardRef`.  To instrument, I am wrapping the RHS of `SwitchWithForwardedRef` with a new `withHeapSwitchAutocapture` HOC.  I'm also setting the component's display name since it's not explicitly defined in `Switch.js` or picked up in by other babel plugins.

## Test Plan
These changes are unit tested in the instrumentor and verified in Detox tests.

## Checklist
- [x] Detox tests pass


instrumentor/test/fixtures/is-switch/output.js
```diff
@@ -1,49 +1,12 @@
-'use strict';
+"use strict";
```
azelaya6, line 1: does it matter if we use double vs single quotes? i noticed we use single quotes in other files.

azelaya6 (approved):
Looks good to me

instrumentor/test/fixtures/is-switch/output.js
bnickel: This doesn't matter.  It's an artifact of the fact `code.js` isn't strict and the output it inserting the declaration.  We don't care how Babel format quotes.

PR #406, opened 2023-09-13
[HEAP-42725] Return the same component each time from `withHeapNavigationWrapper`

bnickel:
## Description

If `Heap.withHeapNavigationWrapper` is called multiple times, it currently returns a new HOC each time.  This means if the app looks like this:

```
const App = () => {

  ...

  const HeapNavigationContainer =
    Heap.withReactNavigationAutotrack(NavigationContainer);

  return (
    <HeapNavigationContainer independent={true}>
      ...
    </HeapNavigationContainer>
  );
};
```

And the app contains a `useEffect` or similar, which caused `App` to re-evaluate, the entire contents of `HeapNavigationContainer` will be considered new and will be recreated, sending the app back to start.

The general fix is to move `Heap.withHeapNavigationWrapper` out of a function but this has come up enough that a code fix is warranted.

This simply caches the output of `withReactNavigationAutotrack` on the wrapped component so it can be retrieved the next time it is called.

## Test Plan

Unit tests were added and executed.  This behavior was vetted in a test app.

It would be possible to write an integration test on this, if someone wants to.

## Checklist
- [x] Detox tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `Heap.withHeapNavigationWrapper` returns the same object when called
+  multiple times, preventing app refreshes when used within a
+  re-evaluated function like `App` with `useEffect`.
+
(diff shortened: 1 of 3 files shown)


MCeley (approved):
👍

PR #412, opened 2023-10-04
[HEAP-49278] Fix capture when React Navigation rootState is null

### LINKED ISSUE #366: The Heap library has crashed with an exception
Our application is built on top of firebase and some of the subscriptions can cause the Navigation stack to redraw. When it happens we may receive undefined in the onStateChange event of react-navigation. That seems properly handled in react-navigation v5. However, the heap navigation wrapper layer crashes in 
NavigationUtil.getActiveRouteNames, as the navigationState is undefined.

callstack of the error:
```
WARN  The Heap library has crashed with an exception. {"message": "Cannot read property 'routes' of undefined", "name": "TypeError", "stack": "TypeError: Cannot read property 'routes' of undefined
    at getActiveRouteNames (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=(hidden)&modulesOnly=false&runModule=true:83071:36)
```

TuurDutoit:
## Description
Fixes #366

In some cases, react-navigation's `getRootState()` can return `undefined`, as documented [here](https://reactnavigation.org/docs/navigation-container/#getrootstate) and [here](https://reactnavigation.org/docs/navigating-without-navigation-prop/#handling-initialization). This was being passed to `getActiveRouteProps()`, causing it to throw: `Cannot read property 'routes' of undefined`

## Test Plan
We applied this change as a patch in our app first, and confirmed it removes the error and successfully sends events. I also added some unit tests for this behaviour here.

## Checklist
- [x] Detox tests pass
    - They failed to builld for me. I don't expect my changes to affect any of those tests though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed issue that prevents events from being sent when the navigator is not initialized
+
 ## [0.22.5] - 2023-09-26
 
 ### Fixed
--- a/js/util/__tests__/navigationUtil.test.ts
+++ b/js/util/__tests__/navigationUtil.test.ts
@@ -1,0 +1,96 @@
+import NavigationUtil from '../navigationUtil';
+
+describe('Navigation utilities', () => {
+  describe('no ref', () => {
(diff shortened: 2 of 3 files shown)


bnickel (approved):
This looks great. Thanks for doing this and for adding the thorough tests!

bnickel:
I plan to include this in a release this week.

PR #415, opened 2023-11-02
[HEAP-49279] Improve error logging

bnickel:
## Description

Our error logging is currently a bit muddled:

- In scenarios where the native module isn't loaded (because it isn't linked or we're in Expo Go), we log a null error on each interaction.
- In scenarios where something is broken, we log that heap has crashed, when in reality that single capture has failed.

This makes the following improvements:

- If the native module is unavailable, a single warning is logged when the file is loaded.  References to `RNHeap` are then null checked and no-op rather than failing.
- SDK methods like `Heap.identify` log their name when they fail, rather than a generic message.
- Other error locations log more specific messages about what failed or "The Heap SDK encountered an error while tracking."

Screenshot 2023-11-01 at 5 00 07 PM

Screenshot 2023-11-01 at 5 02 02 PM

To support this, the following changes were made:
- `bail` and `bailOnError` were renamed to `logError` and `swallowErrors` to match their behaviors.
- `track` functions now log using `swallowErrors` instead of their own `try/catch`.
- Navigation capture failures now just log to the console, not the yellow box, to match other autocaptured events.
- I've fixed up the detox tests _again_ to handle a change in `simctl` and with an aggressive clean function for Android.

## Test Plan
The detox tests were run, along with some manual testing where I renamed native references to simulate failure (NativeModules.RNHeap -> NativeModules.RNHeapX, RNHeap.identify -> RNHeap.identifyX).

## Checklist
- [x] Detox tests pass


MCeley (approved):
Looks good. 👍

PR #416, opened 2023-11-02
[HEAP-49279] Make compatible with `use_frameworks!`

bnickel:
## Description

This adds support for podfiles containing `use_frameworks!`, employing the following three fixes:

1. The podspec explicitly requests its dependencies so framework linking succeeds.
2. The auto-init script now writes the settings to the correct location when using frameworks.
3. The auto-init code now reads the settings from the correct location when using frameworks.

## Test Plan

To support testing, the 0.66 test driver now uses frameworks while the 0.63 driver continues to use static linking.

## Checklist
- [x] Detox tests pass


MCeley (approved):
👍

PR #417, opened 2023-11-03
[HEAP-49350] Release 0.22.6

bnickel:
### Fixed

- Fixed issue that prevents events from being sent when the navigator is not initialized.

- Fixed linking and auto-initialization when using `use_frameworks!`.

### Changed

- Error logging has been refined to surface more timely and precise messages.

## Checklist
- [x] Detox tests pass


MCeley (approved):
👍

